PR #190, opened 2014-09-02
Specify explicit dependencies

forki:
Otherwise we break Paket


YaakovDavis:
Reproducible in VS 2015 RC with github packages.
Tried the following:
`github pezipink/MixinProvider`

Issue #201, opened 2014-09-25
Reference documentation generation with CSharp Extension methods fails


matthid:
When a function and a module is marked with the System.Runtime.CompilerServices.Extension attribute, for example

	[<System.Runtime.CompilerServices.Extension>]
	module Test =
	    [<System.Runtime.CompilerServices.Extension>]
	    let MyExtension (o : obj) = o

 exceptions like the following appear while generating the API reference documentation:

	Error while generating docs for C:\Projects\myproject\build\DLL_WITH_CSHARP_EXTENSIONS.dll: System.AggregateException: One or more errors occurred. ---> Microsoft.FSharp.Compiler.ErrorLogger+ReportedError: Exception of type 'Microsoft.FSharp.Compiler.Tast+InternalUndefinedItemRef' was thrown.
	   at Microsoft.FSharp.Compiler.ErrorLogger.ErrorLoggerExtensions.ErrorLogger.Error[b](ErrorLogger x, Exception exn) in d:\Local\Public\FSharp.Compiler.Service\src\fsharp\ErrorLogger.fs:line 321
	   at Microsoft.FSharp.Compiler.Tast.EntityRef.get_Deref() in d:\Local\Public\FSharp.Compiler.Service\src\fsharp\tast.fs:line 2546
	   at Microsoft.FSharp.Compiler.SourceCodeServices.FSharpEntity.get_CompiledName() in d:\Local\Public\FSharp.Compiler.Service\src\fsharp\vs\Symbols.fs:line 147
	   at FSharp.MetadataFormat.ValueReader.hasAttrib@193.Invoke(FSharpAttribute a) in c:\Tomas\Public\FSharp.Formatting\src\FSharp.MetadataFormat\Main.fs:line 193
	   at Microsoft.FSharp.Collections.SeqModule.Exists[T](FSharpFunc`2 predicate, IEnumerable`1 source)
	   at FSharp.MetadataFormat.ValueReader.buildUsage@314(FSharpMemberFunctionOrValue v, FSharpOption`1 args) in c:\Tomas\Public\FSharp.Formatting\src\FSharp.MetadataFormat\Main.fs:line 315
	   at FSharp.MetadataFormat.ValueReader.buildShortUsage@375.Invoke(Int32 length) in c:\Tomas\Public\FSharp.Formatting\src\FSharp.MetadataFormat\Main.fs:line 375
[...]

PR #208, opened 2014-10-05
C# support (suggestion)

matthid:
This is a suggestion on how to add C# support. 

While now C# assemblies will now be processed we still have the following problems:
- Methods/Constructors will not be shown in the final documentation page, we need support in FSharp.Compiler.Service (maybe we should have this pull request open until the FSharp.Compiler.Service is fixed, as it is not as usefull until this is fixed, if you want to use this you can use matthid/FSharp.Compiler.Service@c5dfd4dd488f6dcd1024b0ed2b564ce9d2d414fa)
- Exception when there are extension methods (#201)
- Obviously line references are not working

The other commits add support for Assembly information in a type (you can add something like: Defined in Assembly "Other.dll" to you type template), see https://github.com/matthid/RazorEngine/blob/develop/doc/templates/reference/type.cshtml#L32. Now the namespaces are sorted (when you have multiple assemblies).

### DIFF
--- a/src/FSharp.MetadataFormat/Main.fs
+++ b/src/FSharp.MetadataFormat/Main.fs
@@ -565,6 +565,7 @@ module Reader =
 
   let getXmlDocSigForMember (memb:FSharpMemberFunctionOrValue) =
     let memberName = 
+      try
         let name = memb.CompiledName.Replace(".ctor", "#ctor")
         let typeGenericParameters =
             memb.EnclosingEntity.GenericParameters |> Seq.mapi (fun num par -> par.Name, sprintf "`%d" num)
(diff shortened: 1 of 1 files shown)


dsyme:
I've taken a look at this.  You should not really be calling ReturnParameter on a non-standard C# Event at all - this gets the type of an event if it is considereed to be like an IEvent-typed property by the F# type system.  But some events are "non-standard" and can't be given a good type as a property, and these must be called using ``add_XYZ`` and ``remove_XYZ``.

That said, I'm adjusting FCS to still give a ReturnParameter with the delegate type of the event in this case. I'm also adding properties EventIsStandard, EventAddMethod and EventRemoveMethod to allow you to be more fine-grained if you need to be.

PR #216, opened 2014-10-13
Fix cache using incomplete key

ovatsus:
This manifested when you had a template.cshtml and a ja\template.cshtml. If I run english docs then japanese docs on F# Data on the same process, the japanese docs would get the english template


matthid:
@ovatsus Can you please check if https://github.com/tpetricek/FSharp.Formatting/pull/274 breaks this again?

PR #224, opened 2014-11-03
Added support for "raw" special tag.

colinbull:
This adds support for a `(*** raw ***)` command. This is effectively the equivalent of a <pre/> html tag, but for markdown.


colinbull:
I'll add it to the docs :).

PR #239, opened 2014-12-30
Relax NuGet package version for Razor and RazorEngine

panesofglass:


tpetricek:
Well, requiring binding redirect would probably negatively affect us (unless I'm missing something?)

 - When generating the docs by running code in F# Interactive, you'd have to add binding redirect to `fsi.exe.config` (wherever that lives)

 - When generating the docs in the usual ProjectScaffold way (via FAKE), you'd have to add the binding redirect somewhere too to make that work (and I don't even know where...?)

panesofglass:
This is now a severe pain point. I am not able to upgrade projects to Paket, and tools like FsReveal, that appear to expect Paket, can't upgrade without me having to tweak their .fsx files.

forki:
Can't we ILMerge these two?

forki:
Also please discuss https://github.com/fsprojects/Paket/issues/515

forki:
The ILMerge will be much easier than fsprojects/Paket#515 and I think it would be more useful in this case.

tpetricek:
I think that lifting the restriction would break common use cases (see my comment above), so we simply cannot do that. What other options do we have? 

 - ILMerge?
 - Update to .NET 4.5 and require more recent version of RazorEngine?
 - Anything else?

tpetricek:
+1 to wait letting Paket to save us :-)

panesofglass:
I'm for ILMerge. This appears to be a restriction special-cased to FSharp.Formatting, which is in all my cases a helper lib. That it prevents me from using Paket on projects is frustrating since the actual project in question should trump helper libs, imho. I'm tempted to fork FSharp.Formatting and publish that fork for use on many of my projects that need newer versions of Razor.

forki:
Please don't fork. I think we can try the ilmerge. I think the paket
solution will need some time and is not the best for this project.

tpetricek:
I actually wouldn't be opposed to switching F# Formatting to .NET 4.5 and using the latest version of RazorEngine. Although I think that would still require referencing a fixed version.

Aside, I guess we do not even need to ILMerge - we could just drop the necessary DLLs into our `lib/net40` folder - but ILMerge sounds nicer (I guess it can make the DLL smaller).

@panesofglass Would you prefer 1. or 2.?

matthid:
While I do not fully understand the problem here: You should be carefull suggesting ILMerging RazorEngine without testing. Chances are that it will simply not work (when the merged assembly has a Location of null). I saw several people reporting this and a workaround is only possiple with 3.5.0 (Again please test before suggesting! I will link the corresponding RazorEngine issue later, currently I'm on my phone).

For switching to .net45: I think a net40 release with RazorEngine 3.5.0 would be nice before (or after) switching (I don't need that personally but maybe there are people that do?). Besides that I never was sure why F# Formatting is still net40 when it only is a build dependency, and it probably is not so much of a problem to have .net45 available on build time as opposed to runtime of the project. Basically you only need net45 for generating the documentation page? 

Would the 3.5.0 release on RazorEngine solve this problem?

forki:
replaced by #246

forki:
@panesofglass did you test the latest version? works pretty well for me.

panesofglass:
I haven't tried as I only just returned from Codemash. I'll try again soon, though I think the solution is both good and should solve my problem. Thank you so much for persevering on a solution!

matthid:
Did we introduce this problem again in the latest versions with "FSharpVSPowerTools.Core"?
Maybe we should split FSharp.Formatting in two nuget packages (one used as build dependency, with all dependencies included, like FAKE) and one to be used as library?

PR #240, opened 2015-01-01
Razor engine update.

matthid:
This reverts commit 1289f612a8d34279add6cffbcdf5eaa1d14c05e4 and updates RazorEngine to version 3.5.0-beta2 (paket doesnt like beta1).
The problem with the previous commit was that upstream projects broke (because of a different dependency).
This time we use the official update of the RazorEngine project.

@forki: I again have a problem with paket. With this change `paket install` will not work:

    Paket failed with:

    Error in resolution.
      Resolved:
        - FAKE 3.12.2
        - RazorEngine 3.5.0-beta2
        - NuGet.CommandLine 2.8.3
        - NUnit 2.6.4
        - NUnit.Runners 2.6.4
        - Microsoft.AspNet.Razor 2.0.30506.0
        - CommandLineParser 1.9.71
        - FSharp.Compiler.Service 0.0.67
      Can't resolve:
        - Microsoft.AspNet.Razor >= 3.0.0
           - from RazorEngine 3.5.0-beta2
     Please try to relax some conditions.

`paket update` worked fine and everything seems to be ok within the project files. Any hint would be very welcome, can we ignore this?
Another thing I noticed: paket seems to add dependencies to the FsLib project files while they do not have any paket.references and they dont need them.

Happy New Year to all who read this :)

### DIFF
--- a/docs/content/literate.fsx
+++ b/docs/content/literate.fsx
@@ -259,4 +259,8 @@ version of the F# compiler:
    `<pre>` tag containing the original source code of the F# Script or Markdown document.
  - `errorHandler` - a function that is used to report errors from the F# compiler 
    (if not specified, errors are printed to the standard output)
+ - `assemblyReferences` - The assemblies to use when compiling Razor templates.
+   Use this parameter if templates fail to compile with mcs on linux or
+   if you need additional references in your templates.
+   (if not specified, we use the currently loaded assemblies)
 *)
--- a/docs/content/metadata.fsx
+++ b/docs/content/metadata.fsx
@@ -91,6 +91,27 @@ module Foo =
 Note that currently our code is not handling `<parameter>` and `<result> tags, this is 
 not so much of a problem given that given that FSharp.Formatting infers the signature via reflection.
 
+
+## Optional parameters
+
+All of the three methods discussed in the previous two sections take a number of optional
+parameters that can be used to tweak how the formatting works:
+
+ - `markDownComments` - specifies if you want to use the markdown parser for in-code comments.
+   With markDownComments enabled there is no support for `<see cref="">` links, so `false` is recommended for C# assemblies.
+   (if not specified, `true` is used)
+ - `typeTemplate` - the templates to be used for normal types (and C# types).
+   (if not specified, `"type.cshtml"` is used)
+ - `moduleTemplate` - the templates to be used for modules.
+   (if not specified, `"module.cshtml"` is used)
+ - `namespaceTemplate` - the templates to be used for namespaces.
+   (if not specified, `"namespaces.cshtml"` is used)
+ - `assemblyReferences` - The assemblies to use when compiling Razor templates.
+   Use this parameter if templates fail to compile with mcs on linux or
+   if you need additional references in your templates.
+   (if not specified, we use the currently loaded assemblies)
+
+
 Work in progress!
 -----------------
 
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -4,7 +4,7 @@ nuget FAKE
 nuget FSharp.Compiler.Service
 nuget CommandLineParser 1.9.71
 nuget Microsoft.AspNet.Razor == 2.0.30506.0
-nuget RazorEngine 3.5.0-beta3
+nuget RazorEngine 3.5.0
 nuget NUnit
 nuget NUnit.Runners
 nuget NuGet.CommandLine
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,8 +3,8 @@ source http://nuget.org/api/v2
 nuget FAKE
 nuget FSharp.Compiler.Service
 nuget CommandLineParser 1.9.71
-nuget Microsoft.AspNet.Razor 2.0.30506.0
-nuget RazorEngine 3.3.0
+nuget Microsoft.AspNet.Razor == 2.0.30506.0
+nuget RazorEngine 3.5.0-beta3
 nuget NUnit
 nuget NUnit.Runners
 nuget NuGet.CommandLine
--- a/paket.lock
+++ b/paket.lock
@@ -2,12 +2,12 @@ NUGET
   remote: http://nuget.org/api/v2
   specs:
     CommandLineParser (1.9.71)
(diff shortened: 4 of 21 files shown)


forki:
Looks like a bug in the new paket install process. Will try to investigate

But paket update and paket restore seems to work.

forki:
> Another thing I noticed: paket seems to add dependencies to the FsLib project files while they do not have any paket.references and they dont need them.

hiw can I reproduce this?

matthid:
That could have been done in an older paket version (one where install still has worked)...

tpetricek:
What is the intended behavior when a project does not have `paket.references`? I never figured out whether that means the project should have *all* references or *none*...

matthid:
I'm also not sure why the build breaks I'll look into that later (locally everything worked). Maybe I should try again with a clean master state or with a fixed paket ...
I can't answer the `paket.reference` question as it seems like `paket` never works for me in the first place (there was already a blocking bug in #209 when I tried it the first time, where we had to add the exactly operator... Is this still required btw?) :)

forki:
I fixed `paket install` in paket 0.20.7 - it was indeed a issue with the == condition.

Regarding missing `paket.referenes` file. Paket will assume the project has no dependencies.

matthid:
Thanks for the quick fix! I will try again later and report the results. So I can now remove the == condition from the `paket.dependencies` file right (the whole indirect reference line)?

forki:
No that's still required. Paket doesn't allow multiple versions of the same NuGet package as dependency.

matthid:
>> Another thing I noticed: paket seems to add dependencies to the FsLib project files while they do not have any paket.references and they dont need them.

> how can I reproduce this?

Just remove the dependencies from `FsLib1.fsproj` and run `paket update --force --hard`, it will re-add while there is no `paket.references` near the `FsLib1.fsproj` file. But there is one in a parent folder (which `paket` seems to use) so maybe that is even intended behavior...

forki:
yes. That's intended behaviour (but maybe not in this case here).

2015-01-01 17:14 GMT+01:00 matthid <notifications@github.com>:

matthid:
Anyone has an idea why the build is now working? It seems like `TargetFrameworkIdentifier` is not defined for executable projects? I can not find any other significant difference between `FSharp.Formatting.CommandTool.fsproj` and for example `FSharp.Literate.fsproj` ...
I always wondered about `TargetFrameworkIdentifier` as it always seems to be `.NETFramework` or unset?

matthid:
Is it OK for now to have a beta package in the dependencies? I uploaded the package myself and I do consider it pretty stable, it's basically only beta for possible backwards compatibility issues.

After merging this pull request we should be able to close https://github.com/fsprojects/ProjectScaffold/issues/119 by adding the optional `assemblyReferences` parameter to all `FSharp.Formatting` calls when we are on mono (see https://github.com/tpetricek/FSharp.Formatting/pull/207 for more details).

tpetricek:
AFAIK, if we have a NuGet dependency on beta version of a package, it means that we won't be able to release F# Formatting as non-beta. That's a bit sad, because our beta would probably be better than our latest non-beta - but I guess we can live with that until RazorEngine reaches non-beta version...

Though, I guess I should still release the current build as non-beta (with the cref fix).

matthid:
Ah I see, so maybe just leave this pull request open until the stable is released (so you can keep doing stable releases on FSharp.Formatting)?

matthid:
RazorEngine 3.5.0 is now released. The update commit is already pushed. 

Can somebody test this on other downstream projects before merging? Are there any downstream projects which use customized cshtml templates (besides me?).

tpetricek:
Yay!

I'll check with Deedle - although that does not use customized template (it used to, before I realized that it was a better default and copied it into F# Formatting :-))

matthid:
Can we merge this (this is a breaking change on some functions as I added parameters to make it run on mono), but the update to 3.5.0 on its own should not be breaking. So I could split those two changes if needed for some reason.

tpetricek:
I think the PR is looking good - thanks for all the work on this! 

Could you please document the additional parameters in the documentation? If I'm correct, one change is in the `Literate.Process[Foo]` functions, which has ["Optional parameters" section](http://tpetricek.github.io/FSharp.Formatting/literate.html#Optional-parameters) in the docs. The MetadataFormat project does not have a similar section yet - if you can add one at the end of [this page](http://tpetricek.github.io/FSharp.Formatting/metadata.html), that would be fantastic (I can fill in the docs for remaining arguments eventually).

If you need more time for this, let me know and I'll merge & release this as it is (we can add docs later).

matthid:
Like [this](https://github.com/tpetricek/FSharp.Formatting/commit/e5db525f13dc48b3b1ba0a970a928e91b72cb9ac)?

tpetricek:
Thanks, this is fantastic! :-)

I'm just releasing this as a new version (2.6.2) and will test it with Deedle later today.

matthid:
Nice Thanks! So you don't think the change is breaking anyone?

tpetricek:
Given that we are now embedding RazorEngine into our NuGet package, I don't think this should be a breaking change (updating to embedded NuGet from non-embedded NuGet would be). In principle, we could break something, but I don't think anyone would call our embedded RazorEngine directly...

... but I certainly want to try this in Deedle docs and see what happens.

matthid:
What I meant was the API change (additional parameter), its too late anyway now that the nuget package is out. What I was trying to say is that 2.7.0 should have been the next version. I hope everything works with Deedle!

I like that we now bundle RazorEngine. I think bundling everything is a good idea for FSharp.Formatting as this enables us to change dependencies without breaking anything...

tpetricek:
Yes, you're right - technically speaking, this is a breaking change. Although AFAIK most people reference this through a script (`generate.fsx`) that is re-compiled when run, and so adding optional parameter will not break things.

tpetricek:
(But I think I'm pretty bad at version numbers - but proper SemVer does not seem al that useful & clear for quickly evolving F# projects.. I'd welcome some F#-friendly guidelines from someone who knows this stuff :-))

PR #243, opened 2015-01-03
Docs around raw command

colinbull:
Here a small amount of documentation about the raw command.


tpetricek:
Thanks! Now it makes perfect sense again :-)

PR #244, opened 2015-01-05
transform code blocks in list correctly

forki:
In FsReveal we often have codeblocks below a list with bullet points.

    ### paket.dependencies
    
    - specifies rules regarding your application's dependencies:
    
        source https://nuget.org/api/v2
        
        nuget Castle.Windsor-log4net ~> 3.2
        nuget NUnit

Result with current FSharp.Formatting:

image

this PR tries to fix this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ tests/Benchmarks/testfiles/*/*.html
 .paket/paket.exe
 release.cmd
 *.orig
+*NCrunch*
--- a/src/FSharp.Markdown/MarkdownParser.fs
+++ b/src/FSharp.Markdown/MarkdownParser.fs
@@ -302,7 +302,7 @@ let (|SkipSomeNumbers|_|) (input:string) =
       Some(rest)
   | _ -> None
 
-/// Recognizes a staring of a list (either 1. or +, *, -).
+/// Recognizes a starting of a list (either 1. or +, *, -).
 /// Returns the rest of the line, together with the indent.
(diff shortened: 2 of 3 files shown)


tpetricek:
I thought you need more spaces when you want to have code block inside a list block. That is, there should be at least 4 spaces relatively to the offset where the usual text starts.

When you try this in [CommonMark](http://spec.commonmark.org/dingus.html) web demo, the following does not create `<pre>`:

    - specifies rules regarding your application's dependencies:

        source https://nuget.org/api/v2

But with two more spaces, it creates a code block:

    - specifies rules regarding your application's dependencies:

          source https://nuget.org/api/v2

I believe we do this already in F# Formatting, but there may be bugs...?

forki:
I tried with tons of spaces

tpetricek:
You're right, it does not work with 4 spaces, although it does with 5 (not sure why):

    """
     - fooo

            a 
       
           """
    |> Markdown.Parse

tpetricek:
I figured out what I found suspicious about this. The problem is that this breaks some nested lists:

     - a
       - b

         c

This is what GitHub does with this:

 - a
   - b

     c

With your change, `c` would be treated as code block...

Now, the way F# Formatting handles lists is already broken in some ways, so I'm not too worried if we break some existing behavior if we make it closer to standard Markdown, but I think this one would be a bit unfortunate.

I'm trying an alternative fix in #245

PR #245, opened 2015-01-05
Alternative attempt to fix #244

tpetricek:


forki:
What about that test case I wrote? Could you please add it, too?

tpetricek:
I don't think your test case should work as it is (it does not work in CommonMark), but I'll add it with 6 spaces. 

I think the idea is that there should be 4 more spaces after the mark that starts the list i.e. `- `. Does that make sense? (I think my fix does that to some point - but not perfectly....)

forki:
I can live with 6 spaces. But I thought the newline (empty line) would end
the list.

tpetricek:
Oh, I see, I misunderstood your point - sorry about that!

tpetricek:
In CommonMark, the following does not end the list:

    - a

        not code

but two empty lines do end the list:

    - a


        not code

forki:
Yeah I tried that too. Didn't work ;-(

tpetricek:
Yeah, that was broken - I finally understand what you were saying :-) sorry for being a bit slow!

I think the last commit should work - you get `<pre>` in `<li>` with 6 spaces or in a separate block after `</ul>` with two newlines.

forki:
image

2 newlines and 6 spaces it works (seems a bit overkill)

tpetricek:
Unless I'm missing something (again), you should need just 2 newlines (to end the list) and 4 spaces (to create a codeblock following the list). The following works on my machine:

    - specified


        source https://nuget.org/api/v2

        nuget Castle.Windsor-log4net -> 3.2     
        nuget NUnit

Alternatively, you could use 6 spaces and 1 newline to create a `<pre>` inside `<li>`:

    - specified

          source https://nuget.org/api/v2

          nuget Castle.Windsor-log4net -> 3.2     
          nuget NUnit

If this does not work, please reopen the issue!

tpetricek:
I might have done something wrong when releasing this yesterday - if 2 newlines + 4 spaces don't work, please try 2.6.0 which will be on NuGet in a sec!

PR #246, opened 2015-01-06
Include RazorEngine and System.Web.Razor in tools - replaces #239

forki:
This will help until we have at least until we have  fsprojects/Paket#515


tpetricek:
I'll release this as 2.6.0 (as we are changing the dependencies again).

PR #247, opened 2015-01-07
Github flavored markdown code-blocks.

matthid:
Initial support for github flavored markup code blocks like

    ```lang
    some code
    ```

see also https://github.com/tpetricek/FSharp.Formatting/issues/181.

### DIFF
--- a/src/FSharp.Literate/Transformations.fs
+++ b/src/FSharp.Literate/Transformations.fs
@@ -20,10 +20,10 @@ module internal Transformations =
   /// to colorize. We skip snippets that specify non-fsharp langauge e.g. [lang=csharp].
   let rec collectCodeSnippets par = seq {
     match par with
-    | CodeBlock({ Code = (String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code)); CodeLanguage = language }) 
-        when (language.IsSome && language.Value <> "fsharp") || (cmds.ContainsKey("lang") && cmds.["lang"] <> "fsharp") -> ()
-    | CodeBlock({ Code = (String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code)) }) 
-    | CodeBlock(Let (dict []) (cmds, { Code = code })) ->
+    | CodeBlock((String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code)), language, _) 
+        when (not (String.IsNullOrWhiteSpace(language)) && language <> "fsharp") || (cmds.ContainsKey("lang") && cmds.["lang"] <> "fsharp") -> ()
+    | CodeBlock((String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code)), _, _) 
+    | CodeBlock(Let (dict []) (cmds, code), _, _) ->
         let modul = 
           match cmds.TryGetValue("module") with
           | true, v -> Some v | _ -> None
@@ -38,8 +38,8 @@ module internal Transformations =
   /// Replace CodeBlock elements with formatted HTML that was processed by the F# snippets tool
   /// (The dictionary argument is a map from original code snippets to formatted HTML snippets.)
   let rec replaceCodeSnippets path (codeLookup:IDictionary<_, _>) = function
-    | CodeBlock { Code = (String.StartsWithWrapped ("[", "]") (ParseCommands cmds, code)); CodeLanguage = language } 
-    | CodeBlock(Let (dict []) (cmds, { Code = code; CodeLanguage = language })) ->
+    | CodeBlock ((String.StartsWithWrapped ("[", "]") (ParseCommands cmds, code)), language, _) 
+    | CodeBlock(Let (dict []) (cmds, code), language, _) ->
         let code = 
             if code.StartsWith("\r\n") then code.Substring(2)
             elif code.StartsWith("\n") then code.Substring(1)
@@ -59,11 +59,10 @@ module internal Transformations =
           else code
         let lang = 
           match language with
-          | Some language -> Some language
-          | None when cmds.ContainsKey("lang") -> Some cmds.["lang"]
-          | _ -> None
-        if (lang.IsSome) && lang.Value <> "fsharp" then 
-          Some (EmbedParagraphs(LanguageTaggedCode(lang.Value, code)))
+          | String.WhiteSpace when cmds.ContainsKey("lang") -> cmds.["lang"]
+          | language -> language
+        if not (String.IsNullOrWhiteSpace(lang)) && lang <> "fsharp" then 
+          Some (EmbedParagraphs(LanguageTaggedCode(lang, code)))
         else
           Some (EmbedParagraphs(FormattedCode(codeLookup.[code])))
 
@@ -274,7 +273,7 @@ module internal Transformations =
           | _ -> None
         match special with 
         | EvalFormat(Some result, _, kind) -> ctx.Evaluator.Value.Format(result, kind)
-        | EvalFormat(None, ref, _) -> [ CodeBlock(CodeBlockInfo.Create ("Could not find reference '" + ref + "'")) ]
+        | EvalFormat(None, ref, _) -> [ CodeBlock("Could not find reference '" + ref + "'", "", "") ]
         | other -> [ EmbedParagraphs(other) ]
 
     // Traverse all other structrues recursively
--- a/src/FSharp.Literate/Transformations.fs
+++ b/src/FSharp.Literate/Transformations.fs
@@ -40,6 +40,10 @@ module internal Transformations =
   let rec replaceCodeSnippets path (codeLookup:IDictionary<_, _>) = function
     | CodeBlock(String.StartsWithWrapped ("[", "]") (ParseCommands cmds, code)) 
     | CodeBlock(Let (dict []) (cmds, code)) ->
+        let code = 
+            if code.StartsWith("\r\n") then code.Substring(2)
+            elif code.StartsWith("\n") then code.Substring(1)
+            else code
         if cmds.ContainsKey("hide") then None else
         let code = 
           if cmds.ContainsKey("file") && cmds.ContainsKey("key") then 
--- a/src/FSharp.Literate/Transformations.fs
+++ b/src/FSharp.Literate/Transformations.fs
@@ -20,10 +20,10 @@ module internal Transformations =
   /// to colorize. We skip snippets that specify non-fsharp langauge e.g. [lang=csharp].
   let rec collectCodeSnippets par = seq {
     match par with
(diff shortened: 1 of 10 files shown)


matthid:
With this change we can use the same markup file as github preview (the readme.md in https://github.com/Antaris/RazorEngine/tree/develop) and as source of the generated documentation (http://antaris.github.io/RazorEngine/) while having syntax highlighting for code snippets in both.

src/FSharp.Literate/Transformations.fs
```diff
@@ -38,8 +38,12 @@ module internal Transformations =
   /// Replace CodeBlock elements with formatted HTML that was processed by the F# snippets tool
   /// (The dictionary argument is a map from original code snippets to formatted HTML snippets.)
   let rec replaceCodeSnippets path (codeLookup:IDictionary<_, _>) = function
-    | CodeBlock(String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code)) 
+    | CodeBlock(String.StartsWithWrapped ("[", "]") (ParseCommands cmds, code)) 
     | CodeBlock(Let (dict []) (cmds, code)) ->
+        let code = 
+            if code.StartsWith("\r\n") then code.Substring(2)
+            elif code.StartsWith("\n") then code.Substring(1)
```
tpetricek: Do we ever expect to get `\r\n` or just `\r` here? I never quite understood newlines in Markdown :-/. Aside from that, it's certainly better to just delete one new line rather than do an arbitrary trim.

src/FSharp.Markdown/HtmlFormatting.fs
```diff
@@ -182,11 +182,18 @@ let rec formatParagraph (ctx:FormattingContext) paragraph =
       ctx.Writer.Write("</p>")
   | HorizontalRule(_) ->
       ctx.Writer.Write("<hr />")
+  | CodeBlock(String.StartsWithWrapped ("[", "]") (ParseCommands cmds, String.TrimStart code))
```
tpetricek: Currently, code formatting is not handled in FSharp.Markdown but only in FSharp.Literate (which adds formatting for F# snippets based on compiler service and also formatting for code snippets). This is done by adding `LiterateParagraph` type ([here](https://github.com/tpetricek/FSharp.Formatting/blob/123131fbf8a00eff4fcf7d3f07a58a972758eca2/src/FSharp.Literate/Document.fs#L27)) and adding special formatting for it ([here](https://github.com/tpetricek/FSharp.Formatting/blob/123131fbf8a00eff4fcf7d3f07a58a972758eca2/src/FSharp.Literate/Transformations.fs#L319)).

Given this, I wonder what is the right structure for adding Github flavored markdown.

In a perfect world, I think the ideal solution would be:

 * Support [fenced code blocks](http://spec.commonmark.org/0.15/#code-fence) as specified by CommonMark in FSharp.Markdown
 * Add support for GitHub-style language tags in FSharp.Literate (so that we can reuse formatting that is already there for other languages and also nice formatting for F#)

Do you think it would make sense to move the language-specific stuff in your PR from FSharp.Formatting to FSharp.Literate? (If that sounds like too much work, we can start with what you have and do the rest later...)

tpetricek:
Nice work - getting this into F# Formatting would be great!!

I added one big comment - wondering if we could separate the core CommonMark snippets from the language-specific formatting (which we're already doing, but in FSharp.Literate - moving stuff there should be OK, because documentation tools always go through FSharp.Literate).

src/FSharp.Literate/Transformations.fs
matthid: I have no idea what the correct way is to remove an initial newline character in this scenario (which belonds to the `[lang=_]` block). The removed trim can lead to formatting errors when using a github flavored codeblock with initial indentation...

src/FSharp.Markdown/HtmlFormatting.fs
matthid: I'm not completely sure if I understand the issue here. I figured that the html output with the `language-*` css classes would be more useful. Why do you consider this code as language specific?

Besides that I think it is no problem removing this part of the PR (the FSharp.Literate stuff doesn't depend on this) if you don't like it.

tpetricek:
Oh, I had another look at this and I think your PR is actually good as it is.  

There is one little thing I'd change - rather than passing the language as a first line `[lang=csharp]`, could you change `CodeBlock` so that it has `string` with the body and another `string` with the "info string" (that is, everything after the fence start)? This is a breaking change, but it is probably the right thing to do according to the [CommonMark specification](http://spec.commonmark.org/0.15/#example-92).

I originally thought that CommonMark does not say anything about language codes, but it says that we should do exactly what you did :-). 

The other thing (future work) is that if you use FSharp.Literate, it actually colorizes code snippets (rather than just generating `<pre>` with language code). So it would be nice to support Github-style code snippets there too (but that is a separate thing).

tpetricek:
So, if you can change this to use `CodeBlock of string * string`, I'm going to merge the PR ASAP. Thanks a lot for looking into this!

matthid:
Thanks for looking into this. I think you are right and we need some cleanup in this area in the future.
The reason I did it the way it is currently, is because it kind of provides a nice fallback. There are (with this PR) two kinds of code-blocks the three backtick code-block (github flavored) and the indentation code block.

The indentation code-block allows you to specify key-value pairs when you put `[key=value]` on the first line. Currently my code converts the three backtick code-block into the `[key=value]` indentation format, so the string of `CodeBlock` is always in the same format and the colorization in FSharp.Literate should be working already.

I don't found the `[key=value]` extension documented anywhere in the specs but I think those are a nice extension to have and more powerful than just providing the code language.

Maybe the better move would be to change it to `CodeBlock of string * (IDictionary<string,string>)`. This would cover both types of code blocks, and it would cover the case of a missing code-language a little better imho (missing `lang` key in the dictionary).

Another nice feature of this conversation is that you can provide additional key-value pairs with:

    ```lang
    [key=value]
    code
    ````

But this PR doesn't create this `language` or `key-value` problem and the `string` in `Codeblock of string` is exactly the same as before. I'm not sure the breaking change is worth it (for now). But I have to agree its definitely not "nice".

Do you agree?

tpetricek:
The big picture is - I think FSharp.Markdown should closely follow the CommonMark spec. In FSharp.Literate, we can do more things (like handle special comments and colorize code ourselves).

So, if we want to make this CommonMark-compliant, we should handle code blocks with info strings as described [here](http://spec.commonmark.org/0.15/#example-93). Given their example:

    ~~~~    ruby startline=3 $%@#$
    def foo(x)
      return 3
    end
    ~~~~~~~

The parser should recognize `ruby` (as the language), the body and the bit `startline=3 $%@#$` (as the rest of the line).

The `[key=value]` syntax is an extension for FSharp.Literate, which I think should be ignored by our Markdown parser.

So, I think that we could use `CodeBlock of string * string` (or possibly with 3 strings to separate language, body & ignored) in FSharp.Markdown (to match CommonMark spec).

When you use FSharp.Literate, we could turn this into a dictionary and perhaps recognize something like this:

    ~~~~    lang key1=value1, key2=value2
    def foo(x)
      return 3
    end
    ~~~~~~~

... but that's a separate thing that should be done only in our extension on top of CommonMark (and I prefer this over supporting `~~~ lang` with `[key=value]` on the next line - the problem with `[key=value]` is that it can sometimes clash with valid code...)

Does this make sense? (This is a tricky thing - so sorry if what I'm writing is confusing - as I said, I'm happy to merge the PR as it is and do other tweaks later)

matthid:
Ah I see, thanks for the explanation it makes sense now. But now I don't understand what you mean by the 3rd string in `CodeBlock` (the rest of the startline ie `startline=3 $%@#$` in your first example?).

Ok I can change that, but for completeness sake and because I'm already in this place now: How do you want to handle the other code blocks now (simple indentation)? Should we try to find the `[lang=lang]` extension and fill up the string in `CodeBlock` or ignore that (I assume now that you want to ignore it)? I think some users might be irritated by FSharp.Markdown results, no matter which way we choose (as currently a lot of documentation uses the `[lang=key]` syntax).

tpetricek:
> what you mean by the 3rd string in CodeBlock

The CommonMark specification says that if you have:

    ~~~ fsharp some random stuff here

Then, the first word (apparently, space separated, so `fsharp`) is the language and the rest should be ignored. We could either find the first word when parsing or when formatting (but we do not want to drop the ignored bit when parsing, because we could use it in FSharp.Literate for other key=value pairs). I don't really have a preference here.

> How do you want to handle the other code blocks now (simple indentation)? Should we try to find the  [lang=lang]  extension and fill up the string in  CodeBlock  or ignore that

The `[lang=fsharp]` commands are currently ignored in FSharp.Markdown, but we *are* handling them in FSharp.Literate (which is what ProjectScaffold runs) - see [the code here](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Transformations.fs#L57). So, if we just ignore them in FSharp.Markdown, than this will continue working - I think this is the best thing to do.

tpetricek:
Also, the source location in FSharp.Literate that I mentioned above is a place where we can try to handle the new Github-style code blocks - so, rather than just parsing the commands in `[...]`, we could also see if there is a language (or other key-value pairs) in the `CodeBlock` itself (the "info string")

matthid:
Thanks again! What do you think of the last commit? I opted for a record instead of a tuple with 3 strings. It should not be too difficult going from here and do the proper implementation (ie. the full spec instead of just triple tick quotes; and adding your proposed `key1=value1, key2=value2` extension).

matthid:
I can squash all commits if you want before merging (left it for now so the comments do not get deleted).

tpetricek:
Although I completely agree that using a record is better than using a DU with `string * string * string`, I'd be tempted to leave it as `string * string * string` for now - mainly because this is how the rest of the API looks at the moment (and so I'd prefer consistency over elegance here).

Once we update this to F# 3.1, we can probably use annotated DUs, which will let us do:

     | CodeBlock of body:string * lang:string * infoString:string

Sorry for asking for that many changes! If you're happy with this change, then I'm ready to merge the PR once it's there (no need to squash commits, IMO).

Thanks again for working on this!

matthid:
Yeah I though that the record might be something that I need to change... Everything should be fine now.

tpetricek:
This is now released in version 2.6.1.

Thanks a lot for sending the pull request. I added a few more tests (for the literate programming parts) and it works perfectly - so once this makes it to the new version of ProjectScaffold, everyone should be able to use it. If you can send a PR to ProjectScaffold to update the version there, that would be great :-)

matthid:
Thanks! Glad if this helps :). Sadly I never used ProjectScaffold until now, so maybe somebody else wants to do that?

forki:
Anything else to do than increasing the version number?

forki:
https://github.com/fsprojects/ProjectScaffold/commit/a601b210d48f9e2ca15091b6edb8eb5403888e0a

tpetricek:
Oh, I didn't even realise how cool Paket is :-)

PR #256, opened 2015-01-25
use SourceCodeClassifier from Visual F# Power Tools Core to colorize F# snippets

vasily-kirichenko:
The result as following, standard (dark) theme:

image

New light theme:

image

Feel free to change the colors of the dark theme as you wish (I make it as close as possible to the colors of VS dark theme (except the color for strings)).

### DIFF
--- a/src/FSharp.CodeFormat/FSharp.CodeFormat.fsproj
+++ b/src/FSharp.CodeFormat/FSharp.CodeFormat.fsproj
@@ -9,7 +9,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>FSharp.CodeFormat</RootNamespace>
     <AssemblyName>FSharp.CodeFormat</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <TargetFrameworkProfile>
     </TargetFrameworkProfile>
     <Name>FSharp.CodeFormat</Name>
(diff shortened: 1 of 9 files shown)


vasily-kirichenko:
Do not merge this please, I will improve it today.

matthid:
Cool, looks very nice!

@tpetricek Would it make sense to bundle all references from now on (in the nuget package)? This way we would not break any `generate.fsx` script by adding/removing references, which was a reason why we were blocked a while before we could update RazorEngine. I think it is well worth the extra download size.

vasily-kirichenko:
I added `Printf` formatters and escaped chars highlighting and now it's ready.

src/FSharp.CodeFormat/FSharp.CodeFormat.fsproj
```diff
@@ -9,7 +9,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>FSharp.CodeFormat</RootNamespace>
     <AssemblyName>FSharp.CodeFormat</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
forki: mhm. that's breaking, right?

vasily-kirichenko: @tpetricek don't mind https://github.com/tpetricek/FSharp.Formatting/issues/255#issuecomment-71334390

tpetricek: @forki Do you think that updating to .NET 4.5 could cause any troubles?

I was worried about this when it came up in the discussion with @matthid, but I don't think it can cause troubles to anyone (the main usage of F# Formatting is through ProjectScaffold and I think that should most likely be fine...)

dungpa:
@vasily-kirichenko Why closing?

forki:
Yes please don't close I don't think it's a big issue

tpetricek:
Please reopen this, I definitely want to get this in!

I think upgrading to .NET 4.5 is probably a good thing to do anyway (also, I was traveling and now I'm on holidays, but I still should have time to review & test & merge this tomorrow).

vasily-kirichenko:
I found a bug (the token next to a string is colored as string). I tried to fix it for an hour and failed. So, I closed this until I have enough time to fix it. The code is not very good indeed partially due to neediness to concatenate subsequent string tokens.

vasily-kirichenko:
Fixed, tests added, a new PR created https://github.com/tpetricek/FSharp.Formatting/pull/257

PR #259, opened 2015-02-01
Add some unit tests for c# and the publiconly flag.

matthid:
It seems like the flag is broken (publiconly=false). My quick investigation shows this is most likely a bug/limitation in FCS. But you are maybe interested in the test cases anyway. Once the problem is fixed in FCS you can remove the Ignore attribute.


tpetricek:
So, the issue is that when you set `publicOnly=false`, it still only shows public members?

matthid:
Exactly. I have added a comment to fsharp/FSharp.Compiler.Service#260.

I have found some other smaller issues regarding the display of properties (when a property implements a set AND a get accessor) and the constructor signature (which is always `unit -> unit`) but I did not manage to properly formulate those issues. Not sure when I manage to send pull requests for those.

tpetricek:
Thanks!

PR #268, opened 2015-02-26
add some search paths to the loading script.

matthid:
please tell me what you think, nuget folders don't need to be named "packages", but we can kind of assume we are in the right folder...

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -80,11 +80,11 @@ Target "Clean" (fun _ ->
 open System.IO
 
 Target "UpdateFsxVersions" (fun _ ->
+    let packages = [ "FSharp.Compiler.Service"; "FSharpVSPowerTools.Core" ]
     let replacements = 
-      [ "packages/FSharp.Compiler.Service.(.*)/lib",
-        sprintf "packages/FSharp.Compiler.Service.%s/lib" (GetPackageVersion "packages" "FSharp.Compiler.Service");
-        "packages/FSharpVSPowerTools.Core.(.*)/lib",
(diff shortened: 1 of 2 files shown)


tpetricek:
I actually think we could drop the paths with `packages` completely. Doing just e.g. `../FSharp.Compiler.Service...` should always work, or not?

Also, I think this change might break the version-update script in the build file: https://github.com/tpetricek/FSharp.Formatting/blob/master/build.fsx#L82. Could you check that?

matthid:
> I actually think we could drop the paths with packages completely. Doing just e.g. ./FSharp.Compiler.Service... should always work, or not?

I have projects where I use both (nuget and paket) to restore packages. It makes sense to have both in this case. It's the question if you want to support such strange environments or only focus on paket within FSharp.Formatting (everything else is the problem of the user).

I think you are right and we should focus (This doesn't really hit anyone as long as all packages belonging to FSharp.Formatting are in either paket or nuget)

> Also, I think this change might break the version-update script in the build file: https://github.com/tpetricek/FSharp.Formatting/blob/master/build.fsx#L82. Could you check that?

Nice catch, I didn't notice that there is such a thing :). It should work now.

matthid:
Do you want me to update RELEASE_NOTES.md here as well? There will be a merge conflict with my other pull request if I do that...

forki:
That's why I do this myself in my projects...

2015-03-03 16:04 GMT+01:00 matthid <notifications@github.com>:

tpetricek:
Thanks, looks good now! I'll update the release notes.

tpetricek:
Perfect! Released in 2.7.5.

PR #270, opened 2015-03-05
FSharp.Formatting fails on projects using EntityFramework.dll.

matthid:
Add support for projects using nuget packages which exist in the system runtime directory (which is preferred over `-I:`).

See https://github.com/fsharp/FSharp.Compiler.Service/issues/313 for details.

This is an example failure: https://travis-ci.org/matthid/Yaaf.Database/builds/53130951
I currently use the `otherflags` parameter as a workaround, however I think it makes sense to integrate the change into FSharp.Formatting directly...

If I did not overlook any side-effects this should do pretty much the same thing as before (without failing on `EntityFramework.dll`)


matthid:
Hm ed9b774 should have worked as well, I don't know why a FSharp.Core.dll gets written into the test directory. But whatever, we support that now...

forki:
top

+1

tpetricek:
OK, I sacrificed a :goat: and released this in v2.7.5 to NuGet. Let's see what happens!

PR #274, opened 2015-03-27
Razor tempfiles cleanup.

### LINKED ISSUE #273: Cleanup RazorEngine files.
Some background: https://github.com/Antaris/RazorEngine/issues/244

Basically RazorEngine was leaking temp files sine 3.5.0. Now I have a fix/workaround in place but it only works when RazorEngine is not used in the default `AppDomain`. So we need to take additional actions here in F# Formatting (because the primary usage is via FSI, which sadly uses the default AppDomain..).

Because spinning up AppDomains in F# scripts will be very difficult (to get a dynamic assembly from one into another AppDomain), I can only see two options:

- Spin up the AppDomain internally in F# Formatting (if we are in the default AppDomain)
  But I guess this is only possible for high level APIs like `MetadataFormat.Generate` (I don't think we can support high order functions)
- Handle this in FAKE: Spin up a new AppDomain within FAKE and run the build-script from there
  Of course we need to also change https://github.com/fsprojects/ProjectScaffold/blob/master/build.template#L199 and spin up another FAKE process instead of calling `fsi.exe`.
  At this point everything should work fine.

IMHO we should opt for 2 (it should be easier to implement as well) and leave F# Formatting unchanged, however I would like to hear what you think.
/cc @tpetricek @forki

matthid:
This kind of fixes  #273 .
Additionally I changed some Razor internals to improve speed (especially if you generate the same documentation for multiple domains, which should now almost be free).
This means we could now per default generate a local documentation as well


ovatsus:
I tried to build this to test on the caching on F# Data, but I have errors while building (upstream/master builds fine):

    Error		The namespace 'Compiler' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	11
    Error		The namespace 'Compiler' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	12
    Error		The namespace 'Compiler' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	14
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	24
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	25
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	26
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	27
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	28
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	29
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	30
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	31
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	32
    Error		The value or constructor 'SourceTokenizer' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	64
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	69
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	71
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	73
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	73
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	73
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	85
    Error		The field, constructor or member 'ColorClass' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	136
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	137
    Error		The field, constructor or member 'ColorClass' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	143
    Error		The namespace or module 'FSharpTokenColorKind' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	144
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	153
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	161
    Error		The type referenced through 'Microsoft.FSharp.Compiler.SourceCodeServices.ToolTipText' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	163
    Error		Type mismatch. Expecting a
        'a -> 'b option    
    but given a
        'a -> obj    
    The type ''a option' does not match the type 'obj'	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	164
    Error		The type referenced through 'Microsoft.FSharp.Compiler.SourceCodeServices.ToolTipText' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	168
    Error		Type mismatch. Expecting a
        'a -> 'b option    
    but given a
        'a -> obj    
    The type ''a option' does not match the type 'obj'	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	169
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	173
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	176
    Error		The field, constructor or member 'TokenName' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	177
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	190
    Error		Type constraint mismatch. The type 
        obj    
    is not compatible with type
        Symbol    
    The type 'obj' is not compatible with the type 'Symbol'	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	190
    Error		The field, constructor or member 'ColorClass' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	192
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	204
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	205
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	214
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	215
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	222
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	225
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	225
    Error		The type referenced through 'Microsoft.FSharp.Compiler.AbstractIL.Internal.Library.Shim.IFileSystem' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	252
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	257
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	258
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	275
    Error		The record label 'OtherOptions' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	281
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	286
    Error		The type referenced through 'Microsoft.FSharp.Compiler.SourceCodeServices.FSharpTokenInfo' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	289
    Error		The type referenced through 'Microsoft.FSharp.Compiler.Range.range' is defined in an assembly that is not referenced. You must add a reference to assembly 'FSharp.Compiler.Service'.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	300
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	345
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	348
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	348
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	348
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	348
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	349
    Error		The namespace or module 'FSharpErrorSeverity' is not defined	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	349
    Error		Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.	FSharp.CodeFormat	C:\Git\FSharp.Formatting\src\FSharp.CodeFormat\CodeFormatAgent.fs	350
    Error		Assembly reference 'C:\Git\FSharp.Formatting\bin\FSharp.CodeFormat.dll' was not found or is invalid	FSharp.Formatting.CommandTool	FSC	1
    Error		Assembly reference 'C:\Git\FSharp.Formatting\bin\FSharp.Literate.dll' was not found or is invalid	FSharp.Formatting.CommandTool	FSC	1
    Error		Assembly reference 'C:\Git\FSharp.Formatting\bin\FSharp.MetadataFormat.dll' was not found or is invalid	FSharp.Formatting.CommandTool	FSC	1

ovatsus:
Ok, problem was just with VS2015, on VS2013 it builds fine (but master did build fine on both)

ovatsus:
Humm, you just made me realize that the template of the current published docs is already broken :/ I'm getting the english template on the japanese docs, and it's not caused by this change, something else before broke it again

matthid:
And with this change it's still not working? Can you link the code which implements this template selection (by language) I want to take a quick look (tomorrow).

ovatsus:
Yes, after this change it's still not working.
https://github.com/fsharp/FSharp.Data/blob/master/docs/tools/generate.fsx#L130

matthid:
The root cause is that https://github.com/fsharp/FSharp.Data/blob/master/docs/tools/generate.fsx#L131 already does recursive processing (you can disable it with `processRecursive = false`). I added it because internally it was already implemented and just not working (so I considered that as bugfix), but knowing that it might break people I added the flag.
Basically your files where generated correctly and than overwritten incorrectly.

Anyway I would recommend you another way of doing this, use the same layout-root for both languages and call your templates "template.cshtml" and "template-ja.cshtml" (`doctemplate` in https://github.com/fsharp/FSharp.Data/blob/master/docs/tools/generate.fsx#L132). You can just enable `processRecursive` and only do your root manually (= select the appropriate template).
This way is easier to manage/debug and you might get some performance if you can share templates across languages (or default back to English for some)

ovatsus:
Thanks, went with the simplest solution for now: https://github.com/fsharp/FSharp.Data/commit/6370c6c339cf35708fa5d308788f2a79926ea17e
Worked fine with this PR, but you should have a look on why doesn't it compile anymore with F# 4.0

matthid:
Did you check if F# Formatting actually compiles without this PR on F# 4.0? Because from the error message I would say that isn't related to this PR at all...

ovatsus:
Yes, it does compile on F# 4.0 without this PR.

matthid:
Quite weird for multiple reasons:
- is F# 4.0 not backwards compatible?
- as `FSharp.CodeFormat` fails to compile and has no dependency on the code I changed here, I would assume the failing has more to do with the updated references `.paket/paket.exe update`. Than with this PR

I currently have no F# 4.0 installation, can you check if it starts failing after `.paket/paket.exe update` and maybe even report that to the F# Compiler people?

ovatsus:
Humm, it now is compiling ok both on VS2013 and VS2015 :/ Don't know how that happened. Never mind then

matthid:
OK, thank you for testing this. So if @tpetricek is fine with the changes this PR is ready.

tpetricek:
Nice work!

I recently noticed that the recursive processing is doing something wrong. Does that work in the version with this PR? What would be the best way to proceed with that? (Should we make `processRecursive=false` the default and update ProjectScaffold, so that it does not go into subfolders explicitly and instead just uses `processRecursive=true`?)

matthid:
I always thought `ProcessDirectory` should already process the directory recursive while keeping its structure and that it didn't do it was a bug. If you disagree feel free to change the default back to `false`. (sorry if that caused any confusion, but the code for recursive processing was already there.. it was just not working properly). But actually that is not part of this PR.

The change was done a while back here: https://github.com/tpetricek/FSharp.Formatting/commit/97c969405af9a831a1d7adae79d402fa615e1656#diff-ac2f189fea923d711033602ea8408a2cL179
Note that the the code was already there, the comment says it would process recursively and the function was marked with `let rec` :)

tpetricek:
Yes, I agree it should be working recursively! 

But I guess all F# projects that use F# formatting include the workaround that calls the function on all the sub-directories, so we'll need to correct this. But I guess not many projects actually use sub-directories. Is this done right in ProjectScaffold?

matthid:
Ah, I agree there is a problem there. We could simply use `processRecursive=false` for now as fix.
If you want to fix projects already forked from ProjectScaffold, I agree that we need to change the default here :(.
But again this doesn't relate to this PR at all. I think globalization is an issue currently here at F# Formatting. 

The solution of ProjectScaffold is a bit overly optimistic and does directly work against how the `ProcessDirectory` API is designed. In fact this is an huge abuse of the API. If you change the template to a non-Razor Template it will break. If F# Formatting takes the full path of the given template into account (like it is supposed to do) it will break. So its doomed to break anyway. I would rather remove this "localization" sooner than later. 

Maybe we should open a new issue and discuss this separately? Or is there and chat-room for F# Formatting where we could discuss possible solutions (as there are several)?

matthid:
After thinking about it, there may be a solution which doesn't disturb users too much and only needs minimal edits in ProjectScaffold (full disclosure: change the template AND the layout-root depending on the language or use only the template-name instead of the full path). We also should ensure that F# Formatting enforces proper usage of the API and I have another Razor caching improvement for this case in the pipeline. Whenever I find some time I will open separate pull requests.

Note that the current PR doesn't change/relate to the situation (as this only does some caching improvements and fixes temporary file cleanup). Should I add `processRecursive=false` here as a quick fix (It would make more sense in a separate PR; I can send that if you want, it's only a single line change though)?

tpetricek:
Let's discuss ProjectScaffold changes as a separate issue - this PR looks good. I'll merge it on Friday (after one nasty deadline :-))

PR #275, opened 2015-03-27
oO, How can F# Formatting itself have no api reference :)

matthid:


misc/templates/template.cshtml
```diff
@@ -25,7 +25,7 @@
       <div class="masthead">
         <ul class="nav nav-pills pull-right">
           <li><a href="http://fsharp.org">fsharp.org</a></li>
-          <li><a href="http://tpetricek.github.io/FSharp.Formatting/"><img height="16" width="16" src="http://fsharp.org/images/thumbs/FSharp.Formatting.png" /> F# Formatting</a></li>
+          <li><a href="http://tpetricek.github.io/FSharp.Formatting/"><img height="16" width="16" src="https://matthid.github.io/Yaaf.AdvancedBuilding/content/img/FSharp.Formatting.png" /> F# Formatting</a></li>
```
matthid: Is there another source for this (with https enabled)? Or are you find with this? This should remove the https warning from the documentation pages.

tpetricek:
I merged this, but then I had a look at `generate.fsx` and I realized that it is a bit unfortunate that we are diverging from the ProjectScaffold version. So, I basically deleted everything there and changed it to do the same thing as what ProjectScaffold does.

If you can review the changes ([all in this commit](https://github.com/tpetricek/FSharp.Formatting/commit/dd03c61e726af20867b1c824e34ab7a6f484fa64)), that would be great!

 - I removed the copied templates (rather than copying them, I made everything point to the `misc` folder)
 - I removed the `html` versions of templates (it all now uses the same `cshtml` template)
 - I copied the `generate.fsx` file from ProjectScaffold

tpetricek:
BTW: @matthid, could you please add your name to the list of authors in `build.fsx` and `generate.fsx`? You did a lot of work on the project and it's shame you're not mentioned there! (Unless, of course, you prefer to be anonymous :-)).

matthid:
Yeah looks good to me (added some thoughts as line notes), I personally don't really like the ProjectScaffold layout (I think the layout is unnecessarily complicated). Normally I use my own (https://github.com/matthid/Yaaf.AdvancedBuilding) which updates with `paket update`, so I'm not always up2date with ProjectScaffold.

Using `.cshtml` everywhere was something I almost wanted to do as well, but I thought it might be good to have an example here that we support these kind of templates as well. On the other side users probably don't look into F# Formatting's build script, so definitely +1 for this. Razor everywhere hehe.

I always thought ProjectScaffolds idea is to get an initial start and then continue however you like (by copying and updating the files). In contrast to stick with it and get updates from it (that's one reason why I use my own solution, see above)
If you think we should always stay up2date with their files we should consider using paket to reference their files instead of making copies (paket can reference github files directly)... This way we would automatically update with `paket update`.

Regarding the authors list, yeah maybe in the next pull request (if I don't forget it)... It doesn't really matter to me that much :)

tpetricek:
Thanks a lot for having a look at this!

PR #279, opened 2015-04-04
remove razor mono workaround (RazorEngine should be able to handle this ...

matthid:
...now, not in every but in the usual FAKE scenario).

See discussion in https://github.com/tpetricek/FSharp.Formatting/commit/dd03c61e726af20867b1c824e34ab7a6f484fa64


tpetricek:
Hmm, looks like it does not :-(

> Processing the file 'docpage' failed
> Source written to: '/tmp/tmp7c270ee7.tmp.cs'
> Compilation errors:
> warning: (0, 0) Assuming assembly reference `FSharp.Core, Version=4.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' matches assembly `FSharp.Core, Version=4.3.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. You may need to supply runtime policy
> error: (0, 0) Assembly `RazorEngine, Version=3.6.4.0, Culture=neutral, PublicKeyToken=9ee697374c7e744a' references `System.Web.Razor, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' which has a higher version number than imported assembly `System.Web.Razor, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
> System.Exception: Generating HTML failed.
> at FSharp.Literate.RazorRender.handleCompile[String] (System.String source, Microsoft.FSharp.Core.FSharpFunc`2 f) [0x00000] in <filename unknown>:0 

If you need more info to figure out what's going on, let me know & I'll try this on my Mac (on Monday).

matthid:
I don't know where System.Web.Razor 2.0.0.0 is coming from, something is wrong. There shouldn't be a 2.0.0.0 anywhere ... So this is not the error I expected here.

matthid:
This seems to be https://github.com/fsharp/FSharp.Compiler.Service/issues/313 again... (there is already a workaround in place for this somewhere else :( ).
That mono decides to put System.Web.Razor in the runtime directory (/usr/lib/mono/4.5) is really a pain as it makes `#I` and `-I` useless, see the related issue.

And actually as the test commit shows it is not even enough to reference the correct assembly, you need to force load it (I don't know if there is a more sane way than what I have done), because otherwise mono again loads the 2.0 version from the runtime directory, even when RazorEngine has an explicit dependency on 3.0. And ironically that explicit dependency is later the reason why the template compilation fails :/

tpetricek:
Nice, you actually got it to work! Hmm, I recall I had to do some force-loading somewhere too... Not quite sure where though....

Which of the changes are actually necessary to make this work? I see you moved some files to the packages folder and added the `dll` there - is that necessary (I'd prefer not to have checked-in things in the `packages` folder, if possible...)? What is the minimal change to get this to work correctly?

matthid:
IIRC I moved it there because there was a bootstrapping problem (the relative paths were wrong). So as long as we don't want to add `../packages`(and we still need to move it one directory) we need to place it in the packages folder.
we could of course use another (duplicate) script for bootstrapping, or copy it to the correct destination via FAKE?

tpetricek:
Would it work if we used the following?

    #I "lib/net40"
    #r "System.Web.Razor.dll" 
    // Force load 
    if (typeof<System.Web.Razor.ParserResults>.Assembly.GetName().Version.Major <= 2) then 
        failwith "Wrong System.Web.Razor Version loaded!" 

... or does the `#r` reference need to point exactly to the file? (Because, this can handle both the case when the DLL is in the sub-directory and when it is in the same directory as the FSX file). I had the impression that F# Compiler Service loads a wrong version, but F# Interactive loads the right version - provided that we force-load it - so I think this might actually work...?

matthid:
That's the painful part: No it will not work: 
#288
https://travis-ci.org/tpetricek/FSharp.Formatting/builds/57761414#L2982

I actually spoke with the mono guys in the gitter chat about this and they told me to write into the mailing list, which I will do eventually. The problem is that they have a `System.Web.Razor.dll` with version 2 in the runtime directory (`/usr/lib/mono/4.5`). This basically means any `-r:System.Web.Razor.dll` is redirected to it, no matter what.

I think there might be another mono or F# bug in it because the explicit reference to Razor-3 (from RazorEngine) is clearly ignored in runtime. I expect at least a `TypeLoadException` with an inner: "RazorEngine or one of its dependencies could not be found". Instead it actually works until we try to compile a template (that's why I introduced the early failure here).

PR #280, opened 2015-04-05
Even more global caching, now even with different layout-roots.

matthid:
This is part of what was discussed in #274, after seeing how the layout-roots parameter is used for different localization. 

Now there is no more reason against using the layout-roots parameter for this (as long as you do not specify the same template via full-path as ProjectScaffold is doing it).

### DIFF
--- a/src/Common/Razor.fs
+++ b/src/Common/Razor.fs
@@ -36,6 +36,36 @@ open RazorEngine.Configuration
 open RazorEngine.Compilation.ReferenceResolver
 
 /// [omit]
+module PathHelper =
+  let private isUnix =
+    match Environment.OSVersion.Platform with
+    | PlatformID.Unix -> true
+    | PlatformID.MacOSX -> true
+    | _ -> false
+  let normalizePath p =
+    let fullPath = Path.GetFullPath(p).TrimEnd(Path.AltDirectorySeparatorChar).TrimEnd(Path.DirectorySeparatorChar)
+    if isUnix then fullPath else fullPath.ToLowerInvariant()
+      
+/// [omit]
+type PathTemplateKey(name, path, t, context) =
+  let path = PathHelper.normalizePath path
+  member private x.Path = path
+  interface ITemplateKey with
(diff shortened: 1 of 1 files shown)


tpetricek:
Hmm, are you saying that this is better way of doing things, but it will break most ProjectScaffold-based projects if they update to newer version of F# Formatting?

(That would be very unfortunate! Or am I misunderstanding something?)

matthid:
This PR should be harmless. But you are correct for my other PR (with the additional commit).
It should only break for people who have forked since the bug was introduced (and only their localization will be hit, everything will be english until they fix the bug). The only alternative I see is to live with the bug.

tpetricek:
Oh, this only affects localizations... now I finally understand this (thanks for making that clear)! 

Then we should be good - I think only F# Compiler Service and F# Data have translations, so I think we just need to fix:

 - https://github.com/fsharp/FSharp.Data/blob/master/docs/tools/generate.fsx
 - https://github.com/fsharp/FSharp.Compiler.Service/blob/master/docs/tools/generate.ja.fsx
 - https://github.com/fsharp/FSharp.Compiler.Service/blob/master/docs/tools/generate.fsx

Is that correct?

matthid:
Uh, Ok I looked into this again (especially since those projects had their localization long before I introduced the bug) and I think I got the impact wrong.
I now think no project is hit by that because ProjectScaffold uses a forward template `docpage.cshtml` and it doesn't really matter if it is localized or not, because it only forwards to `template.cshtml` which will in turn be localized correctly... So https://github.com/fsprojects/ProjectScaffold/pull/155 wasn't strictly required (but it's still a good change to make IMHO).

This change (#281 to be exact) only hits people relying on the bug, which appeared in Sep 2014. Everything what worked before that date should still work after applying this (or work again if it broke).

(Sorry for the confusion)

tpetricek:
Oh, right - thanks for digging deeper! 

I'll release a new version with the changes then. And I agree that the change in ProjectScaffold is a good one anyway.

PR #281, opened 2015-04-05
Use template path if it is rooted.

matthid:
This is a bugfix for the case of an empty layout-root and a full-path template.
This also enables the use of a full-path template outside of the layout-root folders.

This does break ProjectScaffold however as they use a full-path template but depend on the bug that this full path is overwritten by the layout-root folder.
Fix is https://github.com/fsprojects/ProjectScaffold/pull/155

### DIFF
--- a/src/Common/Razor.fs
+++ b/src/Common/Razor.fs
@@ -194,11 +194,16 @@ module RazorEngineCache =
     engine
 
 /// [omit]
-type RazorRender(layoutRoots, namespaces, templateName:string, ?references : string list) =
-  let templateName =
-    if templateName.EndsWith(".cshtml") then
-        templateName.Substring(0, templateName.Length - 7)
-    else templateName
+type RazorRender(layoutRoots, namespaces, template:string, ?references : string list) =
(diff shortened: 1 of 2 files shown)


tpetricek:
Sorry, I'm also a bit confused here... 

From the source code, it looks like the new version handles both full paths and relative paths (using the `Path.IsPathRooted` check) - but ProjectScaffold used a full path - so, why is that a breaking change?

(Again, I'm trying to be a bit careful here - the problem is not just ProjectScaffold, but what worries me is that all projects based on it would break if they updated to new version of F# Formatting, which they sooner or later will....)

matthid:
It really depends what you expect when you use a full path template, currently we completely ignore the path and we even throw when layout-roots is empty. IMHO a full path should always be considered, and actually I think it was like that pre RazorEngine 3.5, but lost in the update process...

As mentioned in the other PR only localization is hit, and the way ProjectScaffold is setup everything will be english, so it probably doesn't hit a lot of users (could be wrong though)

Issue #251, opened 2015-01-15
Customize parsed markup output.


matthid:
Sometimes you want to be able to change the output of the html generation: https://github.com/matthid/FSharp.Formatting/tree/prismjs_support

But a lot of this is hard-coded: https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Markdown/HtmlFormatting.fs

So I'm suggesting to replace this code with RazorEngine templates. This way it would be easier for users to replace the templates. Would you accept this change? 
If there is still interest in a RazorEngine alternative we would want to keep the generation code as I don't think we can do the same with other template engines.

I can already think of some problems:
 - `*.cshtml` templates are written in C# so the code will not be "nice and clean" because of F# interop
 - It complicates things.
 - There will a lot of new small templates, which all needs to be compiled which will lead to longer run-times (cold start)

But there would be some benefits:
 - Users can edit single templates without the need to fork, edit and build FSharp.Formatting.
 - One can use their own strategy to colorize code-blocks.
 - Other outputs than Latex and Html are possible (by just using your own templates)
 - no direct mathjax dependency for formulas

What do you think? I'm thinking about this some time now, but I'm not sure if it's worth the effort.

tpetricek:
Hmm, this is a tricky question. I would really like to keep the "Markdown" part of F# Formatting as simple as possible - currently, we're using Razor in the "Literate" programming part and in the "MetadataFormat" project, but not in the pure Markdown.

The fact that we parse Markdown into an AST means that you can quite easily customize the output just by doing an AST transformation (the [documentation has an example](http://tpetricek.github.io/FSharp.Formatting/markdown.html#Processing-document-recursively)) and e.g. replace all code blocks with HTML blocks (formatted using your preferred tool) without touching the rest of the document. I think this approach would work just fine in the scenario that you linked above (primjs_support).

That said, this could certainly be implemented as an extension on top of F# Formatting (it could use F# Formatting for parsing, but then implements new rendering). So, I think you can do this as a separate project - and I'd be happy to link to it, in case people want to customize formatting using Razor.

matthid:
I think this would work well for "Markdown" and "Literate", but for "MetadataFormat" I would need an additional API to replace the `Markdown.WriteHtml` calls in https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.MetadataFormat/Main.fs#L490-L497. 
[...]

PR #282, opened 2015-04-06
Add a public API to process a customized LiterateDocument.

matthid:
Now that Literate if internal my fix for #251 doesn't work anymore :(.
See also https://github.com/tpetricek/FSharp.Formatting/commit/dd03c61e726af20867b1c824e34ab7a6f484fa64.

Do you like another optional parameter for that? Alternatively we could use a03fcfb.

### DIFF
--- a/docs/content/literate.fsx
+++ b/docs/content/literate.fsx
@@ -268,4 +268,7 @@ version of the F# compiler:
    Use this parameter if templates fail to compile with `mcs` on Linux or Mac or
    if you need additional references in your templates
    (if not specified, we use the currently loaded assemblies).
+ - `customizeDocument` - Allows you to customize the document before writing it 
+   to the output file. This gives you the opportunity to use your own
+   code formatting code, for example to support syntax highlighting for another language. 
 *)
--- a/src/FSharp.Literate/Main.fs
+++ b/src/FSharp.Literate/Main.fs
@@ -51,7 +51,12 @@ type Literate private () =
       if references <> Some true then doc
       else Transformations.generateReferences doc
     doc              
-      
+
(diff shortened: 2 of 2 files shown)


tpetricek:
Oh, hmm - I guess we don't need to make it internal. I did that just to cleanup the public API reference, but we can just `[omit]` the module.

For individual document, you should be able to use `ParseScriptFile`, then do the transform and then call `WriteHtml`. Or is there some reason why this wouldn't work?

Now, I see that this cannot be done for `ProcessDirectory`. I'm not a very big fan of complex callbacks ([see this blog post](http://tomasp.net/blog/2015/library-frameworks/index.html)), so I'd probably prefer an API where you can call `ParseDirectory` (which returns a list of files with parsed documents) and then `WriteHtml` taking the list and writing all documents. Would that let you solve the problem?

matthid:
`WriteHtml` is very limited and it doesn't do the same processing that https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Formatting.fs#L88 does.

Sure there are a lot of ways to solve this, I think the callback is reasonably simple though (basically `LiterateDocument` -> `LiterateDocument`).
Besides that I first experimented with a03fcfb (an alternative solution) which would also solve my problem (and is a unification of `WriteHtml` and `WriteLatex` actually). What do you think about that API?

tpetricek:
The `ProcessDocument` looks like a nice addition to me.

If that solves your problem too, I'd be much happier to accept that (and yes, you're right - the `WriteHtml` and `WriteLatex` functions wouldn't work because they omit the transformations...). I think adding `ProcessDocument` would be good even if we had the callback, actually...

PR #285, opened 2015-04-07
Update LaTeX colors using CSS Light theme

### LINKED ISSUE #278: LaTeX formatting of F# code snippets is broken
The PR to use F# PowerTools formatting added HTML formatting for the additional types of tokens, but **not** for LaTeX. I added [workaround for this here](https://github.com/tpetricek/FSharp.Formatting/commit/2a1af08710228f7cb879c00a2b55af93586a4bd6), but it would be great to cover the additional cases properly (we do not need to make the LaTeX template fancier, but we should map the new tokens to the closest existing kind of token).

@vasily-kirichenko Could you have a look at this?

dungpa:
Close #278.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ misc/literate/output/content/
 .paket/paket.exe
 release.cmd
 *.orig
+*.aux
+*.dvi
+*.out
+*.synctex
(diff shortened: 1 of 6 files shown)


tpetricek:
Fantastic, thank you!

PR #286, opened 2015-04-07
updated typical literate setup

Indy9000:
Updated the text explaining the typical setup.


tpetricek:
Wonderful, thanks a lot for the contribution :-)!

Indy9000:
Pleasure :) 
I still don't see the changes on the webpage though.

tpetricek:
The changes do not appear on the page automatically – only when we release a new version.

(Which I did just yesterday, so they should be visible now.)

 

Thanks again!

PR #289, opened 2015-04-09
Add an API to process a customized LiterateDocument.

matthid:
as discussed in #282 

You can merge both, select one or select none and remote the `internal` from `Templating`. Either solution is fine with me.

One thing that concerns me is that this style of API doesn't work well with F# composition. For example `ProcessMarkdown` and `ProcessScriptFile` should be able to easily reuse this IMHO.
Maybe we should sometimes in the future refactor all the optional parameters into a record?

One concrete problem how I use it here: https://github.com/matthid/ircdotnet/blob/ci_docs_and_build/generateDocs.fsx#L64
Notice how I need a `ctx` object to customize the transformation depending on the output format (which is what you always want to do with this API). But if you depend only on the `doc` and `output` parameters (`format` is optional) you basically need to depend on F# Formatting internals (eg. using the extension to detect the format). So maybe I should make `format` non-optional (in contrast to their counter-parts).
What do you think (as you seem to like the API more than I do)?

### DIFF
--- a/src/FSharp.Literate/Main.fs
+++ b/src/FSharp.Literate/Main.fs
@@ -133,6 +133,13 @@ type Literate private () =
   // Processing functions that handle templating etc.
   // ------------------------------------------------------------------------------------
 
+  /// Process the given literate document
+  static member ProcessDocument
+    ( doc, output, ?templateFile, ?format, ?prefix, ?lineNumbers, ?includeSource, ?generateAnchors,
+      ?replacements, ?layoutRoots, ?assemblyReferences) =
+    let ctx = formattingContext templateFile format prefix lineNumbers includeSource generateAnchors replacements layoutRoots
+    Templating.processFile assemblyReferences doc output ctx
+
   /// Process Markdown document
   static member ProcessMarkdown
     ( input, ?templateFile, ?output, ?format, ?formatAgent, ?prefix, ?compilerOptions, 


tpetricek:
I merged both - I prefer the `ProcessDocument` API, but I think the one with callback can be useful too.

As for the API design, I think that what F# Formatting provides currently is a reasonable simple-to-use very high level API. But it would make sense to clan up the internals a bit and make them public (and documented) to give an alternative and more composable alternative.

Something like:

    Literate.parseMarkdown fileName
    |> Literate.evaluateSnippets ctx
    |> Literate.replaceEvaluatedSnippets ctx
    |> Literate.format Html

This would have additional benefit of actually clarifying and documenting the various transformation steps (which, at the moment, are somewhat ad-hoc and pretty hard to follow). If you feel like doing this, that would be pretty awesome :-)

matthid:
Yes I completely agree. 

In fact, if no-one else complains about the now internal API we have a lot options now without breaking anyone.
Additionally we might want to collect those suggestions in a markup file ("future directions") or in separate issues so other collaborators can take them as well (as I don't know when and if I have time for it).
IIRC there is also the suggestion in the room to refactor "/src/FSharp.MetadataFormat/Main.fs" which we discussed in a previous PR/issue (and which I also agree is highly needed).

The problem with new issues is that we already have a lot ...

:+1: for cleaning up the issues list before even sending the comment :)

PR #292, opened 2015-04-12
Remove Tests.sln

matthid:
Ok maybe I'm doing this wrong but the normal workflow for FSharp.Formatting is like this:

- Open Tests.sln and add failing unit test
- (B) Open .sln and add the feature or bugfix
- Build (otherwise it wont work and you run the tests on an outdated build)
- Switch back to Tests.sln (rebuild!), run the test again
- When the test succeeds go to (F) or back to (B)
- (F) Close Test.sln and .sln

Which is kind of cumbersome, this PR simplifies the workflow to

- Open .sln
- Write unit test
- Run, Write Code, Run, Write Code ...
- Close .sln

This also fixes the problem of running the tests on an outdated build (which happens all the time to me). But again I might be doing things wrong :).


tpetricek:
I don't really have a good reason for keeping the two separate. I was just used to it :-). Perhaps the only technical issue is that VS sometimes locks the generated XML doc file (for no reason) and so you may need to restart VS more often.

In any case, the merged solution works for me too, so I'm happy to use this instead.

Issue #287, opened 2015-04-08
FSharp.MetadataFormat.Reader.readAssembly: Argument Exception: An item with the same key has already been added


zealisland:
It's caused by code like this

/// Function Foo!
abstract member Foo: int->uint
/// Empty function for signature
default x.Foo a = 
   ()

The generated XML has two member entries that have the same key

PR #293, opened 2015-04-12
fix #229 and #287

### LINKED ISSUE #229: Key already exists exception when parsing an assembly
When there's a documented instance member and interface implementation with the same name - ````Name```` - in the snippet below, MetadataFormat.Generate fails with an exception, because there're two identical member names in a .xml file (````<member name="P:TestFormatting.Test.Name">````) and no references are generated. As a workaround one of the comments has to be removed...

    type ITest = abstract member Name : string

    type Test (name) =
        /// instance comment
        member x.Name = name

        interface ITest with
            /// interface comment
            member x.Name = name  

Exception:
````System.ArgumentException: An element with the same key already exists in the dictionary.
  at System.Collections.Generic.Dictionary`2[System.String,System.Xml.Linq.XElement].Add (System.String key, System.Xml.Linq.XElement value) [0x00000] in <filename unknown>:0 
  at FSharp.MetadataFormat.Reader.readAssembly (Microsoft.FSharp.Compiler.SourceCodeServices.FSharpAssembly assembly, Boolean publicOnly, System.String xmlFile, Microsoft.FSharp.Core.FSharpOption`1 sourceFolderRepo, Microsoft.FSharp.Core.FSharpFunc`2 urlRangeHighlight, Boolean markDownComments) [0x00000] in <filename unknown>:0````

### PR DESCRIPTION
**Pull request by matthid**
As the comment explains we can pretty much only take the last:
The entries in the xml file are:

```xml
<member name="P:FsLib.Test_Issue229.Name">
<summary>
 interface comment
</summary>
</member>
<member name="P:FsLib.Test_Issue229.Name">
<summary>
 instance comment
</summary>
</member>
```

and

```xml
<member name="M:FsLib.Test_Issue287.Foo(System.Int32)">
<summary>
 Empty function for signature
</summary>
</member>
<member name="M:FsLib.Test_Issue287.Foo(System.Int32)">
<summary>
 Function Foo!
</summary>
</member>
```

So the underlying problem is that what the F# compiler gives us is just not enough information to be useful for us. Maybe someone can look more deeply into this and *maybe* we can (ab)use the ordering somehow. However this change should be enough to fix #229 and #287 .

### DIFF
--- a/src/FSharp.MetadataFormat/Main.fs
+++ b/src/FSharp.MetadataFormat/Main.fs
@@ -946,8 +946,11 @@ module Reader =
         // an easy way to detect where "value" is coming from, because the entries
         // are completely identical. 
         // We just take the last here because it is the easiest to implement.
+        // Additionally we log a warning just in case this is an issue in the future.
         // See https://github.com/tpetricek/FSharp.Formatting/issues/229
         // and https://github.com/tpetricek/FSharp.Formatting/issues/287
+        if xmlMemberMap.ContainsKey key then 
+          Log.logf "Warning: Duplicate documentation for '%s', one will be ignored!" key
         xmlMemberMap.[key] <- value
 
     let ctx = ReadingContext.Create(publicOnly, assemblyName, xmlMemberMap, sourceFolderRepo, urlRangeHighlight, markDownComments, urlMap)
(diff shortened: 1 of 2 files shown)


tpetricek:
This looks perfectly reasonable. Would it make sense to open an issue for this in the http://github.com/Microsoft/Visualfsharp/ repo?

PR #298, opened 2015-04-18
WIP: Support for generating WikiMedia files from Markdown

colinbull:
Most companies I work at use media wiki as a reference point for documentation. This PR adds support for generating the wiki media documents. 

_Note:_ I think there are folder structure conventions for media wiki to import the files, but I'm not sure at the minute it just copies the structure of the HTML generator.


tpetricek:
Looks cool - I think the CodeFormat isn't going to cause issues, but I'll have a more detailed look a bit later (I unfortunately need to run too :-))

matthid:
@colinbull Is this finished or still WIP? And if it is: Can you explain whats missing (so we can decide if it makes sense to already merge this)?

colinbull:
Hi,

I believe this generates the correct markup (based on some very simple
tests). However I just emit the files in the same folder structure as the
html output. I don't believe this is the required format for wiki-media.
However we have since migrated from this to static web pages. So can't
verify this directly.

On Mon, Sep 21, 2015 at 6:39 PM, Matthias Dittrich <notifications@github.com

matthid:
Closing due to inactivity or no more interest to keep this up2date. Please comment if someone has interest to update this.

Issue #261, opened 2015-02-03
FS0192: internal error: null: GetTRefType


vasily-kirichenko:
I updated `FSharp.Formatting` package in `FSharp.Configuration` to 2.7.1 and tried to generate docs. It crashes with 
```
FS0192: internal error: null: GetTRefType
``` 
on https://github.com/fsprojects/FSharp.Configuration/blob/master/docs/tools/generate.fsx#L95-L102 (I mean if I evaluate the script function-by-function, incrementally, the error appears when I evaluate the highlighted lines). 

If I revert to 2.6.2, the docs are generated OK.

What I can found is only this blog post http://blog.mavnn.co.uk/strange-errors-from-the-f-compiler/index.html/ my @mavnn

tpetricek:
You need to add `#r` to reference the VS PowerTools dependency. With [this change](https://github.com/fsprojects/ProjectScaffold/commit/29be5a01404a755755c90f1812c1c6b456cdf4d9) it should work.

forki:
That will break a lot of consumers, right? Can't you ilmerge?

vasily-kirichenko:
Yes, it's helped. BTW, the snippets are too colorful for my taste. Patterns are too bright and I don't think we need to highlight operators now. My idea was to mimic Visual Studio C# editor colors + add a new color for F# DUs/active patterns. Don't get me wrong, I naturally love new coloring :) but the following snippet is a bit too much

![image](https://cloud.githubusercontent.com/assets/873919/6025960/918ff9fc-abe7-11e4-91b7-1275e8a185bd.png)

Maybe removing patterns colorization or choosing a different color for them would make better results?

tpetricek:
I suppose IL merge would make sense - @vasily-kirichenko do you have any thoughts on that? Can I get one of you guys to send a PR for this :-)?

As for the too bright colours - I agree. I experimented a bit with a version that had orange keywords and various yellow-ish tones for strings/formatters/escaped and various blue-ish tones for modules/types/DUs. I think it looked better because similar things had similar colours. I'll send a PR with that.

tpetricek:
How about this:

![colours](https://cloud.githubusercontent.com/assets/485413/6030555/8ed20726-abf8-11e4-89a6-5be119f578aa.png)

It uses yellow-ish colours for strings etc. and aqua-ish colours for identifiers etc.; I made comments gray to make it a bit less colourful.

matthid:
[...]

PR #302, opened 2015-04-21
add back RazorEngine.dll

matthid:
I must have lost my mind when removing this. Good news is that this only seems to lead to problems with the fsf 2.9 release when using the RazorEngine classes directly in the build scripts. The bad news is ProjectScaffold like projects do this because of the mono workaround :(

And I don't know the root cause of https://github.com/tpetricek/FSharp.Formatting/issues/261 but similar things could happen?

I don't really know how we could reliable prevent something like this in the future. Maybe it is viable to add a step to the build script which clones ProjectScaffold, replaces FSharp.Formatting in the packages folder and tries to generate the docs? This could however lead to problems with dependencies and unnecessary build failures. Maybe paket could help us there by providing a way to install a nuget package file directly (/cc @forki)? This would help in other situations as well if you quickly want to test a nuget package before actually releasing it.

### DIFF
--- a/packages/FSharp.Formatting/FSharp.Formatting.fsx
+++ b/packages/FSharp.Formatting/FSharp.Formatting.fsx
@@ -11,10 +11,12 @@ if (typeof<System.Web.Razor.ParserResults>.Assembly.GetName().Version.Major <= 2
 // Standard NuGet locations
 #I "../FSharp.Compiler.Service.0.0.87/lib/net45"
 #I "../FSharpVSPowerTools.Core.1.8.0/lib/net45"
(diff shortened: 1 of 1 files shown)


matthid:
Ok here is why I removed those lines: Strictly speaking we don't need the `#I` lines because we bundle `RazorEngine.dll` and it will be available in `lib/net40`. Removing the `#r` line slipped through somehow.

tpetricek:
Thanks! The bug #261 is pretty mysterious to me - in any case, just releasing an updated version with this fix.

Issue #303, opened 2015-04-24
Fix FSharp.Formatting dependency hell.


matthid:
This is a continuation of https://github.com/tpetricek/FSharp.Formatting/issues/277, https://github.com/tpetricek/FSharp.Formatting/issues/261 and https://github.com/tpetricek/FSharp.Formatting/pull/239 to finally get rid of those.

IMHO we should only go down the ILMerge road if we have literally no other option, using ILMerge on libraries will probably increase the pain for library users which happen to have the same dependencies (FCS comes into mind). I think the notes of http://fsharp.github.io/2015/04/18/fsharp-core-notes.html apply basically for all other dependencies as well (TL;DR: libraries should not bundle and should not ILMerge dependencies).

The good thing is there are at least two other solutions:
- Create some kind of wrapper `.exe` like `FAKE.exe` which wraps the execution of our `generate.fsx`. With this we can basically configure our assembly redirects (for that exe) or even consider ILMerge (if we still want to do that for some reason). We could even have our own DSL and ask @forki how he would do it if he would re-do FAKE :). For the start if would probably be good to support existing scripts if possible.
- Do what you already did: Fix the versions and additionally bundle `all` dependencies. This is probably the lower cost solution.

Note how both options contradict my initial statement (we started contradicting it since bundling RazorEngine). So additionally we should split out our use cases in two seperate NuGet packages 
- FSharp.Formatting <- used as it is currently as a build time helper (with one of the above options) with no additional dependency (everything is bundled to "just work" in build scripts; with one of the options from above)
- FSharp.Formatting.Library <- used in case someone wants to use FSharp.Formatting as library, for example the markdown parser (bundles no dependencies but has nuget dependencies). Users of the library need to handle assembly redirection, just like for any other library.

_Sidenote (Paket):
[...]

PR #305, opened 2015-04-24
Simplify FCS interaction code by using Yaaf.FSharp.Scripting.

matthid:
*EDIT*

As discussed below, this PR now references the source code of https://github.com/matthid/Yaaf.FSharp.Scripting and fixes several bugs:

 * When specifying an `FsiEvalutator` we currently have the nasty side effect of changing the current directory, this will follow up with a vast number of different exceptions (depending on which paths are relative and which are full)

* The current "fsi"-object workaround did not work on mono.

* When something goes wrong in FCS we get improved exceptions.

* Our own documentation generation has now the same environment, this makes sure we detect problems (like the above) sooner here at FSharp.Formatting (#306).

At the same time the code is simplified.

---------------------------
*Original (continuation of #303)*

We cannot see the "real" error, because FSharp.Formatting is swallowing it...

For the FsLab issue:

The root cause seems to be in https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Evaluator.fs#L130 which pretty much expects all libraries to be available near FSharp.Literate.dll. As it is searching for compile-time assemblies, I'm now wondering why it is actually working on windows. I think we will know more when we have the exact error message...
While I'm looking for the exact error this can already run through the CI build in the mean time...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.suo
 bin/
 obj/
+paket-files/
 nuget/*.nupkg
 docs/output/
 literate/outputs/*
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ docs/output/
 literate/outputs/*
 packages/*
 !packages/FSharp.Formatting
+packages/FSharp.Formatting/lib
 *.VisualState.xml
 TestResults.xml
 TestResult.xml
--- a/FSharp.Formatting.sln
+++ b/FSharp.Formatting.sln
@@ -7,7 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "project", "project", "{194B
 	ProjectSection(SolutionItems) = preProject
 		build.fsx = build.fsx
 		nuget\FSharp.Formatting.CommandTool.nuspec = nuget\FSharp.Formatting.CommandTool.nuspec
-		src\FSharp.Formatting.fsx = src\FSharp.Formatting.fsx
+		packages\FSharp.Formatting\FSharp.Formatting.fsx = packages\FSharp.Formatting\FSharp.Formatting.fsx
 		nuget\FSharp.Formatting.nuspec = nuget\FSharp.Formatting.nuspec
 		paket.dependencies = paket.dependencies
 		RELEASE_NOTES.md = RELEASE_NOTES.md
@@ -39,9 +39,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{312E452A-1
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "sidebyside", "sidebyside", "{561B2DD7-455E-4464-B585-C8768C633EFC}"
 	ProjectSection(SolutionItems) = preProject
-		docs\content\sidebyside\extensions.md = docs\content\sidebyside\extensions.md
-		docs\content\sidebyside\markdown.md = docs\content\sidebyside\markdown.md
-		docs\content\sidebyside\script.fsx = docs\content\sidebyside\script.fsx
+		docs\content\sidebyside\sideextensions.md = docs\content\sidebyside\sideextensions.md
+		docs\content\sidebyside\sidemarkdown.md = docs\content\sidebyside\sidemarkdown.md
+		docs\content\sidebyside\sidescript.fsx = docs\content\sidebyside\sidescript.fsx
 	EndProjectSection
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tools", "tools", "{BEE70727-0C9E-42DE-A65B-5CB8F4EF1820}"
--- a/docs/tools/generate.fsx
+++ b/docs/tools/generate.fsx
@@ -31,11 +31,20 @@ open Fake
 open System.IO
 open Fake.FileHelper
 
-// The following lines are F# Formatting only bootstrapping...
+// This is needed to bootstrap ourself (make sure we have the same environment while building as our users) ...
+// If you came here from the nuspec file add your file.
+// If you add files here to make the CI happy add those files to the .nuspec file as well
+// TODO: INSTEAD build the nuspec file before generating the documentation and extract it...
 ensureDirectory (__SOURCE_DIRECTORY__ @@ "../../packages/FSharp.Formatting/lib/net40")
-File.Copy
-  ( __SOURCE_DIRECTORY__ @@ "../../packages/Microsoft.AspNet.Razor/lib/net45/System.Web.Razor.dll", 
-    __SOURCE_DIRECTORY__ @@ "../../packages/FSharp.Formatting/lib/net40/System.Web.Razor.dll", true)
+let buildFiles = [ "CSharpFormat.dll"; "FSharp.CodeFormat.dll"; "FSharp.Literate.dll"
+                   "FSharp.Markdown.dll"; "FSharp.MetadataFormat.dll"; "RazorEngine.dll";
+                   "System.Web.Razor.dll" ]
+let bundledFiles =
+  buildFiles
+  |> List.map (fun f -> 
+      __SOURCE_DIRECTORY__ @@ sprintf "../../bin/%s" f, 
+      __SOURCE_DIRECTORY__ @@ sprintf "../../packages/FSharp.Formatting/lib/net40/%s" f)
+for source, dest in bundledFiles do File.Copy(source, dest, true)
 #load "../../packages/FSharp.Formatting/FSharp.Formatting.fsx"
 
 open FSharp.Literate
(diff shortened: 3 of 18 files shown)


matthid:
Ok thats the message after this patch:

```
Error:
System.Exception: FSI failed:
        Error:

/home/matthid/projects/FsLab.Templates/input.fsx(1,1): error FS0078: Unable to find the file 'FSharpVSPowerTools.Core.dll' in any of
 /usr/lib64/mono/4.5
 /home/matthid/projects/FsLab.Templates/packages/FSharp.Formatting/lib/net40
 /home/matthid/projects/FsLab.Templates
 /home/matthid/projects/FsLab.Templates/packages/FAKE/tools/

        Out:  ---> System.Exception: Evaluation failed ---> Microsoft.FSharp.Compiler.Build+FileNameNotResolved: Exception of type 'Microsoft.FSharp.Compiler.Build+FileNameNotResolved' was thrown.
  at Microsoft.FSharp.Compiler.Build+TcConfig.ResolveLibWithDirectories (Microsoft.FSharp.Compiler.CcuLoadFailureAction ccuLoadFaulureAction, Microsoft.FSharp.Compiler.AssemblyReference r) [0x00000] in <filename unknown>:0
  at Microsoft.FSharp.Compiler.Build+TcImports.TryResolveAssemblyReference (Microsoft.FSharp.Compiler.AssemblyReference assemblyReference, Microsoft.FSharp.Compiler.ResolveAssemblyReferenceMode mode) [0x00000] in <filename unknown>:0
  at Microsoft.FSharp.Compiler.Build.tryFile@4408 (Microsoft.FSharp.Compiler.TcImports tcImports, range m, System.String speculativeFileName) [0x00000] in <filename unknown>:0
  at Microsoft.FSharp.Compiler.Build+TcImports.ImplicitLoadIfAllowed (range m, System.String assemblyName, Boolean lookupOnly) [0x00000] in <filename unknown>:0
  at Microsoft.FSharp.Compiler.Build+TcImports.FindCcuInfo (range m, System.String assemblyName, Boolean lookupOnly) [0x00000] in <filename unknown>:0
...
```

It's not like super nice but definitely better than before. But still: I have no idea why it is working on windows :(.

matthid:
But yes after looking at this, bundling as proposed in #303 would help here as well...

matthid:
@tpetricek : I'm giving up on the FsLab/FCS issue, I guess there is code somewhere in FCS to consider currently loaded assemblies when resolving (compile) assemblies, but I couldn't find it. Maybe this is even just another FCS issue?

We can work around this particular issue within FSharp.Formatting quite easily by just adding additional "#I" and "#r" statements not only for the currently executing, but for all assemblies we need. 

Or we consider this as by design (IE tell users of FSharp.Formatting that we expect our dependencies to live in the same folder as "FSharp.Formatting.dll").

tpetricek:
Ah! Thanks a lot for looking into this, I think I now understand what's going on (looks like I did break something with a recent change :-( ...)

I basically added [the broken piece of code](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Evaluator.fs#L130) to avoid dependency on `FSharp.Compiler.Interactive.Settings.dll` which was major source of pain... So, I changed the code to initialize dummy `fsi` object, but that introduced this new issue.

I think the safest option is just to eval some ugly inlined string with the definition of `fsi` object at the beginning.

BTW: I used this to get better error info: http://fssnip.net/qF

tpetricek:
Well, this was definitely my mistake - after fixing this (I'll send a PR), we won't need to have all the libraries in the same folder, so perhaps that also reduces the urgency of #303 - but I think the discussion there is still valid and it is probably a good direction (but we can perhaps wait a bit more to see if other people have any thoughts on that...)

matthid:
> to avoid dependency on FSharp.Compiler.Interactive.Settings.dll

This is really a PIA, that's one of the reason why I have extracted the scripting use case in https://github.com/matthid/Yaaf.FSharp.Scripting for a scripting API with all those little bugfixes like nice error messages (this PR) or fixing the `FSharp.Compiler.Interactive.Settings.dll` thing.

Currently we have those fixes spread out in several locations:
* https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Evaluator.fs#L127
* https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/FSIHelper.fs#L122

To show two examples :)

> we won't need to have all the libraries in the same folder

Ah, so we don't actually need to load ourself into the interactive we are hosting? That should definitely fix this as well :)

However I still feel like we should add this (or something similar), because it will give people using FSharp.Literate exactly the error message they would get when running in FSI.exe (and it will simplify our lives at well when people report errors or when we encounter them)

> BTW: I used this to get better error info: http://fssnip.net/qF

Oo never noticed this, thanks!

> but we can perhaps wait a bit more to see if other people have any thoughts on that...

Yes to get feedback was exactly the intention of #303 ...

tpetricek:
Yaaf.FSharp.Scripting looks very nice! I would probably prefer to avoid another DLL dependency, but if we could import something like that as a few source files via Paket, that might be nice (and I suspect FAKE could do the same).

The way I understand the F# Compiler Interactive Service thing is that:

 1. `FsiEvaluationSession.GetDefaultConfiguration(new InteractiveSettings(), false)` where `InteractiveSettings` is your type - this gives you `fsi` object, but it requires `FSharp.Compiler.Interactive.Settings.dll`.

 2. `FsiEvaluationSession.GetDefaultConfiguration()` this does not give you `fsi` object and does not require `F.C.I.S.dll` (FAKE does this and as a result you cannot just `#load "FsLab.fsx"` in a FAKE script, because FsLab calls `fsi.AddPrinter`...). I did a workaround ([part 1](https://github.com/fslaborg/FsLab.Templates/blob/journal/build.sh#L11) & [part 2](https://github.com/fslaborg/FsLab/blob/master/src/FsLab.fsx#L40)) for this.

 3. `FsiEvaluationSession.GetDefaultConfiguration()` and then `EvalInteraction("let fsi = ...")` like what [we are doing here](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Evaluator.fs#L134) is a way to get `fsi` object without requiring `F.C.I.S.dll`. (But we need to do it without having to reference ourselves - I thought that was clever, but it breaks things. Basically, having the implementation in a big string is probably OK.)

Having all the tweaks in some file that we could reference via Paket would be nice (and we could send PR to FAKE to use it too.) It would be nice to add (3) to FAKE too - I could then get rid of my dirty workarounds for (2)...

matthid:
Very nice idea! I will try to refactor the features in easy to include source files!

> FsiEvaluationSession.GetDefaultConfiguration(new InteractiveSettings(), false) where InteractiveSettings is your type - this gives you fsi object, but it requires FSharp.Compiler.Interactive.Settings.dll.

Actually I experienced the reverse when working these things out in the past: https://github.com/fsharp/FSharp.Compiler.Service/blob/89d936ca760b790f7741a0bab9122002a792810c/src/fsharp/fsi/fsi.fs#L2705 The `useauxlib` parameter means adding a reference to `F.C.I.S.dll`. It's definitely hard, confusing and impossible to get it perfectly right, I learned that the hard way :)

> But we need to do it without having to reference ourselves - I thought that was clever, but it breaks things. 

The funny thing is we could do this in `Yaaf.FSharp.Scripting.dll` but only when it's a separate dll without and additional references :). Not when it's included via source files.

> FsiEvaluationSession.GetDefaultConfiguration()

https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/fsi/fsi.fs#L2855 it literally uses `(obj, false)`
But FAKE uses `FsiEvaluationSession.Create(` directly: https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/FSIHelper.fs#L122

Bah I hope at least of that is correct :) . I'll report back once I have refactored the files.

matthid:
Uh, I think some of my "magic" is not working when using as script file :(

matthid:
> FsiEvaluationSession.GetDefaultConfiguration(new InteractiveSettings(), false)

Did you mean `true` here? Then everything makes sense now...
So you are saying that we cannot make the `fsi` object work properly with `false` (ie no `F.S.I.S.dll` reference) but we can hook in our own implementation to not fail immediately (even when scripts now produce potentially different output)?

Ok done, I re-used this PR (the title is still true).
This is my magic string by the way: https://github.com/matthid/Yaaf.FSharp.Scripting/blob/develop/src/source/Yaaf.FSharp.Scripting/YaafFSharpScripting.fs#L316.
But now after re-reading your post it probably doesn't really matter what implementation we use (any dummy implementation would probably do the same)...

matthid:
Added another simplification.

matthid:
OK finally everything seems to work as it should. A lot of commits but actually only a very few amount of changes to review (details in the edited start comment).

tpetricek:
I wanted to be sure that relative paths are still handled correctly, so I [added a test for that](https://github.com/tpetricek/FSharp.Formatting/commit/8c056e1fd9e62151485171ee61058f72c639ac51) and... everything worked perfectly :-)

In fact, I think it is actually even redirecting the work to the `fsi` object (which is doing a few unexpected things in FsLab - like showing charts - but that's okay - this is the right thing to do :-))

matthid:
Thanks for the testing and the release!

> everything worked perfectly :-)

That was exactly the intention: Make F# scripting work as easily as possible without the need to care about all this FCS nuances/bugs/argument stuff.

But something (which looks related to this change) is broken now :( , looking into it...

> wanted to be sure that relative paths are still handled correctly,

After noticing this I'm also wondering why `FsiEvaluator` is not used by default? I have planned to enable it for all my projects (now, after this has been merged & released). At least we could enable it in `ProjectScaffold` so you can use those ["enable script output"](http://tpetricek.github.io/FSharp.Formatting/evaluation.html) features out-of-the-box?

> In fact, I think it is actually even redirecting the work to the fsi object (which is doing a few unexpected things in FsLab - like showing charts - but that's okay - this is the right thing to do :-))

Thanks for this, I did some testing myself but I wasn't sure if everything works as it should when some advanced stuff is done with it :)

Everything is done with reflection so it may be a bit slow, but I don't think anyone is calling the `fsi` object like crazy anyway (only for initial setup).

Btw FAKE does now also contain this, so you may be able to remove some workarounds you mentioned above (please report if something breaks)

tpetricek:
I think we could enable it for ProjectScaffold - can you send a PR there? We initially thought it was a bit experimental and perhaps we don't want to run things on people's machine when building docs :-) but I think it is mature and people expect this now...

matthid:
Yeah, but first I'll need figure out why something broke with the update...

PR #306, opened 2015-04-26
Dont Merge

matthid:
This checks if we can reproduce the mono bug fixed in #305, these changes will be integrated there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ docs/output/
 literate/outputs/*
 packages/*
 !packages/FSharp.Formatting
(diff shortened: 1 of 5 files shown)


matthid:
Interesting, I can see the error [here](https://travis-ci.org/tpetricek/FSharp.Formatting/builds/60092970#L3689), but the build doesn't fail.

matthid:
Ok, the build succeeds, because the current directory is exactly the same for all documentation pages.

matthid:
I could reproduce both errors.

PR #308, opened 2015-04-29
use the otherFlags parameter.

matthid:
This is what #305 broke. Actually we probably should even remove that parameter now. If there is a problem to workaround we should probably add it to FSharp.Scripting...

For now that would be a breaking change, but maybe something to note for a major update. Is it possible to obsolete this optional parameter?


forki:
whatever helps to get it working again.

matthid:
@forki Do you mix this with the fake issue? Or did this actually hit you (I though no-one is actually using that "otherFlags" parameter besides me)? If this hit you I would be interested for what you use that parameter...

forki:
woah. sorry. Yep I mixed it. shame on me ;-)

tpetricek:
Released in v2.9.4.

PR #309, opened 2015-05-02
Add API to invalidate templates (RazorEngine 3.7.0).

matthid:
Additionally I used this new API to provide `./build.sh WatchDocs` via https://github.com/fsharp/FAKE/pull/766 and RazorEngine 3.7.0

This is really a cool idea taken from FAKE (see linked PR), but improved in performance by not running a new process every time something changes. 

You can now just run `build WatchDocs`, and every documentation edit will trigger a documentation re-creation.

While implementing this I noticed that we have two Razor caches for `FSharp.Literate` and `FSharp.MetadataFormat` which is not ideal. We probably want to have only a single cache (extract the cache to a new F# project?)...

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -226,6 +226,10 @@ Target "NuGet" (fun _ ->
 Target "GenerateDocs" (fun _ ->
     if not <| executeFSIWithArgs "docs/tools" "generate.fsx" ["--define:RELEASE"; "--define:REFERENCE"; "--define:HELP"] [] then
       failwith "generating reference documentation failed")
+      
+Target "WatchDocs" (fun _ ->
+    if not <| executeFSIWithArgs "docs/tools" "generate.fsx" ["--define:WATCH"] [] then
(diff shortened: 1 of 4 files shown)


tpetricek:
Looks good, thanks - and sorry for the delay!

BTW: I think that the watcher in FAKE already has some queue to process the updates in batches. Do we need the one here? (If yes, can we instead change FAKE so that it lets us pass a timeout?)

matthid:
I actually use the FAKE queue, but I think it has a multi threading problem which I did run into:

When a handler is long running it will execute the next even if the previous one is still running...
That's why I added another queue here (which guarantees only one handler is running) ... But no idea if this is still true.

tpetricek:
Hmm, I see - it would be much nicer if we could just use FAKE queue. I think the feature in FAKE is still quite new and not many people rely on it, so perhaps we could change it there? (especially if it makes things better!)

PR #314, opened 2015-05-22
Add Module and Namespace in nested module and type pages

### LINKED ISSUE #53: Render the namespace in detail pages
Hi,

I just noticed there is no namespace information on the details pages. This might be a problem if someone googles the page:

image

Cheers,
Steffen

LimoWanKenobi:
Adds module and namespace information to the "TypeInfo" and "ModuleInfo" types. 
This information is used to add namespace name and links to the parent module in type.cshtml and module.cshtml templates.
Adds test case to review that templates can be generated with correct information.
This will fix #53

### DIFF
--- a/misc/templates/reference/module.cshtml
+++ b/misc/templates/reference/module.cshtml
@@ -35,7 +35,7 @@
 <h1>@Model.Module.Name</h1>
 <p>
   <span>Namespace: @Model.Namespace.Name</span><br />
-  @if(FSharpOption<Module>.get_IsSome(@Model.ParentModule))
+  @if(Model.HasParentModule)
   {
   <span>Parent Module: <a href="@(Model.ParentModule.Value.UrlName).html">@Model.ParentModule.Value.Name</a></span>
   }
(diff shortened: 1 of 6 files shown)


tpetricek:
This looks like a very useful addition, thanks for doing this! 

The only thing I would perhaps do a bit differently is to expose the additional information in a way that makes it a bit easier to use in the C# code in the template, so that you do not have to write things like:

    if (FSharpOption<Module>.get_IsSome(@Model.ParentModule)) ...

Wouldn't the following work instead?

    if (Model.ParentModule.IsSome) ...

I think this should actually work - if so, could you please simplify the template?

(if this doesn't work, then I'm not sure what would be better - I guess we could add `HasParentModule` or even return `null` - woohoo!)

LimoWanKenobi:
Yep, while adding the link to the template it didn't feel right doing `if (FSharpOption<Module>.get_IsSome(@Model.ParentModule)) ...`, I will review if I can do something better than returning null.

LimoWanKenobi:
Definitely `if (Model.ParentModule.IsSome) ...` is not possible.
I have added `HasParentModule` property to `TypeInfo` and `ModuleInfo` records, and I have added the `Exists` extension  method to the `Module option` following @ldfallas http://langexplr.blogspot.co.at/2008/06/using-f-option-types-in-c.html 

I prefer the Exists extension method. If you agree I would remove the HasParentModule property.

tpetricek:
Thanks, this is a nice trick :-).

PR #316, opened 2015-05-28
Fixed a collection of simple text typos in the documentation.

Vextasy:


tpetricek:
I always suspected I'm bad at English articles, but now I have a proof :-)
Thanks a lot for sending such a thorough correction!

PR #319, opened 2015-06-08
Added code tag in literate transformation of a code block.

kfuglsang:
This is related to https://github.com/fsprojects/FsReveal/pull/74

Code blocks are now transformed such that they are within a code-tag. This allows reveal.js/highlight.js to correctly highlight the code.


tpetricek:
I did one more change and sent you a PR: https://github.com/kfuglsang/FSharp.Formatting/pull/1

Unfortunately, adding `<code>` there breaks some of the styling:

err

Perhaps it is that having `\n` after `<pre>` does not add a new line, but having `<code>\n` after `<pre>` does? I'm not entirely sure - if you could have a look at this, that would be great. 

This is exactly the subtle differences that HTML/CSS changes often cause us :-(

tpetricek:
You can generate the documentation by opening `generate.fsx` in editor and running it step by step. There is an `#if TESTING` directive that lets you (when running this by hand), change `website` URL so that the CSS loads properly on your local machine from the file system.

kfuglsang:
Looks good now with the latest commit ;-)

forki:
exciting

tpetricek:
Sounds good - worked on my machine too :-). I'm just releasing a new version!

It would be wonderful if you could check how the CSS affects the look of some of the existing slides that are out there (like [some from my talks repo](http://github.com/tpetricek/Talks) e.g. [this one](https://github.com/tpetricek/Talks/tree/master/2015/end-to-end-web/usa)). Just to make sure this has no unintended consequences :-). But if it works on the standard samples in FsReveal, than that's good enough. That's why we have `paket.lock` :-)

tpetricek:
Just used this for the updated R provider docs and it seems to be working there too :) nice!

PR #320, opened 2015-06-14
Add class to specify if FSharp.Formatting was able to highlight a code snippet

kfuglsang:
This is needed for fixing https://github.com/fsprojects/FsReveal/pull/76 which requires us to know if FSharp.Formatting was able to highlight a code snippet or not.


forki:
This LGTM and I'd really want to use it in my talk. So @tpetricek sorry for releasing this - I hope it's OK for you.

tpetricek:
Cool with me, thanks for looking into this!

tpetricek:
Also, feel free to do the same thing in the future if you need to :-) with Paket, it is so easy to fix a dependency and issues even if something breaks, so I feel it is much better to move faster if you're working on something in other projects!

PR #322, opened 2015-06-22
Fix HTML escaping of code blocks with unknown languages

### LINKED ISSUE #321: FSharp.Literate does not escape HTML entities in code blocks with unknown language
When using e.g. `FSharp.Literate.ProcessDirectory` to generate HTML from Markdown, **code blocks with unknown languages** (e.g. `batchfile`) receive no escaping of special HTML characters (&, <, >).

This issues seems to only affect code blocks with languages that are unknown to `CSharpFormat.SyntaxHighlighter.FormatCode`. Code blocks with known languages and inline code blocks do receive correct escaping.

Contents:
* **Reproduction steps**
* **Details / Root cause**
* **Possible solution**

## Reproduction steps
In practice, that means that:
```
[lang=batchfile]
program --argument <string>
```
results in the following HTML:
```html
<code>program --argument <string></code>
```
instead of:
```html
<code>program --argument &lt;string&gt;
```

## Details / Root cause
When a code snippet is to be formatted, the following steps occur (only relevant parts are included):

1. [`Transformations.replaceSpecialCodes`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Transformations.fs#L350) calls [`SyntaxHighlighter.FormatCode`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SyntaxHighlighter.cs#L47) to turn the raw code into formatted HTML.
2. [`SyntaxHighlighter.FormatCode`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SyntaxHighlighter.cs#L95) looks up the correct source formatter `sf` for the specified language.
  1. *If the language is known*: `sf.FormatCode(string)` is called. The [default implementation](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SourceFormat.cs#L134) calls the private function [`SourceFormat.FormatCode`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SourceFormat.cs#L187), which, besides other things,  replaces special HTML characters (&, <, >) with the appropriate HTML entities.
  2. *If the language is not known*: The raw code is returned without changes. **No encoding of special HTML characters is applied.**
3. [`Transformations.replaceSpecialCodes`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Transformations.fs#L350) writes the result of `SyntaxHighlighter.FormatCode` into the output (that is, either the HTML output from the formatter, or the unescaped code).

## Possible solution
The solution for this problem is rather simple: Either [`SyntaxHighlighter.FormatCode`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SyntaxHighlighter.cs#L95) or [`Transformations.replaceSpecialCodes`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.Literate/Transformations.fs#L350) calls [`SyntaxHighlighter.FormatCode`](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/CSharpFormat/SyntaxHighlighter.cs#L47) should apply escaping of special HTML characters even if the language is not known.

In my opinion, the change should be applied to  `SyntaxHighlighter`, because the safe escaping of the output should not depend on whether the language is known or not.

On the other hand, if `CSharpFormat.SyntaxHighlighter` is considered to be a public API, the change might break some clients outside of `FSharp.Formatting` that expect `SyntaxHighlighter.FormatCode` to return the unescaped input if the language is not known (e.g. because they already use a workaround for the issue described here). But it is also likely that other clients produce invalid HTML because they do not employ a workaround.

cr7pt0gr4ph7:
Fixes #321. Also extended two existing unit tests to cover this scenario.

The pull request also contains some changes to the `.fsproj` files that enable them to be compiled on machines that have Visual Studio Express (without the F# plugin) and the Visual F# Tools 3.1 or 3.0 installed. (This was necessary to make it compile on my machine). All previous compilation environments are supported without changes.

In general, the `CSharpFormat` library might need a cleanup w.r.t functionality that is both broken and unused by the current code (see, for example, the implementation of `lineNumbers` and `alternate` in `SourceFormat.FormatCode(string, bool, bool, bool)`).

### DIFF
--- a/src/CSharpFormat/CodeFormat.cs
+++ b/src/CSharpFormat/CodeFormat.cs
@@ -101,18 +101,29 @@ namespace Manoli.Utils.CSharpFormat
 			get { return true; }
 		}
 
+		// Match group numbers for the Regex built in the constructor
+		private const int COMMENT_GROUP = 1;
+		private const int STRING_LITERAL_GROUP = 2;
+		private const int PREPROCESSOR_KEYWORD_GROUP = 3;
+		private const int KEYWORD_GROUP = 4;
+		private const int OPERATOR_GROUP = 5;
+
+		/// <summary>
+		/// A regular expression that should never match anything.
+		/// </summary>
+		private const string IMPOSSIBLE_MATCH_REGEX = "(?!.*)_{37}(?<!.*)";
+
 		/// <summary/>
 		protected CodeFormat()
 		{
 			string regKeyword = BuildRegex(Keywords);
 			string regPreproc = BuildRegex(Preprocessors);
 			string regOps = BuildRegex(Operators);
-			if (regOps.Length == 0)
-				regOps = "(?!.*)_{37}(?<!.*)"; //use something quite impossible...
-			if (regPreproc.Length == 0)
-				regPreproc = "(?!.*)_{37}(?<!.*)"; //use something quite impossible...
-			
-			//build a master regex with capturing groups
+			if (regOps.Length == 0) regOps = IMPOSSIBLE_MATCH_REGEX;
+			if (regPreproc.Length == 0) regPreproc = IMPOSSIBLE_MATCH_REGEX;
+
+			// Build a master regex with capturing groups.
+			// Note that the group numbers must with the constants COMMENT_GROUP, OPERATOR_GROUP...!
 			StringBuilder regAll = new StringBuilder();
 			regAll.Append("(");
 			regAll.Append(CommentRegEx);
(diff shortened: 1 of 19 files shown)


tpetricek:
Looks good to me, thanks a lot for the contribution! 

The project file changes should be fine. We used to have odd project files to make compilation work with older version of Mono, but I think we don't need that anymore.

tpetricek:
I just published this as version 2.9.9 :-).

PR #324, opened 2015-06-29
Add test for parser crash on invalid string

tpetricek:
The `categorizedSpans` function (defined [https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.CodeFormat/CodeFormatAgent.fs#L298](here)) crashes when the input string contains incorrect string. Our current way of handling invalid F# code is that we still return something, so it  would be nice to keep this behavior even in this case.

(This was triggered by the F# Snippet: http://fssnip.net/snippet/hH/0, which previously parsed - in some way).

/cc @vasily-kirichenko (or perhaps @dungpa) Could you look into this? This seems to be happening in code that calls Visual F# PowerTools...


matthid:
Seems to be fixed now (just rebased against the latest changs). Either PowerTools Update or our changes.

PR #329, opened 2015-07-08
Adds strong tag to namespace and Parent Module labels

LimoWanKenobi:
Adds strong tag to namespace and Parent Module labels. It will look nicer.


tpetricek:
I guess the whitespace changes are caused by Atom?

(Not a problem, just checking - I'm using Atom too and it does this to me as well)

LimoWanKenobi:
Yes, I deactivated that plugin from Atom and had to copy my changes to a new branch for my other PR #330. It was too much noise.

tpetricek:
Thank you. Closing this one then.

Issue #295, opened 2015-04-17
Handle <parameter> etc for XML comments


zealisland:
In http://tpetricek.github.io/FSharp.Formatting/metadata.html, it mentions that 

_Note that currently our code is not handling \<parameter\> and ` tags, this is not so much of a problem given that given that FSharp.Formatting infers the signature via reflection._

Just want to check whether this is on the roadmap? If so, when will \<parameter\> etc be supported?

Thanks!

tpetricek:
We'd be happy to add support for this. 

If you're interested in contributing, this is something that can be handled in the `readXmlComment` function [here in the source code](https://github.com/tpetricek/FSharp.Formatting/blob/master/src/FSharp.MetadataFormat/Main.fs#L503). I think that all that needs to be done is to change it to recognisze the `<parameter>` tag and turn it into `<ul>` HTML element with `<li>` elements for all parameters (and append this to the generated string). So it should be a nice issue for a new contributor :-).

(There is also a project `csharpSupport` in `tests` folder, so this is a place where you can add a sample documented method, which can then be processed and checked by the tests.)

matthid:
I think this has been merged an released with #331.

PR #331, opened 2015-07-21
[wip] support for all recommended xml tags

ademar:
The purpose of this PR is to bring support for all recommended tags (ref #295)

https://msdn.microsoft.com/en-us/library/vstudio/5ast78ax(v=vs.100).aspx

TODO: unit tests


tpetricek:
Very cool, it will be great to have this!

matthid:
Yeah I think this is very useful, thanks! Can you rebase your changes against the latest changes (ideally against the `fix_extension_methods` branch)?

ademar:
Guys, I have rebased this PR against master.

I have disabled this test https://github.com/tpetricek/FSharp.Formatting/commit/4f4c3c288f01abc97eedb4976bd5a8cf6e5b8b37

In our use case, `markDownComments = false`, we want to render all comments regardless of whether or not they are enclosed by XML tags.

It would be nice if this gets merged, I could use it over here https://github.com/SuaveIO/suave/pull/239

matthid:
There are multiple reasons why I think we should support special commands, even when `markDownComments = false`:

- It should be as simple as for the markdown case to exclude some comments from the website (while still having them in the IDE).
- Problem is that sometimes `FSharp.Compiler.Service` throws errors on some language constructs, and `[omit]` is an easy workaround for such cases.

So it was definitely "by-design" that I called https://github.com/tpetricek/FSharp.Formatting/pull/331/files#diff-e2212151e1f9b6e7e085820d2a6b71caL652 (`findCommands`) in both cases.

matthid:
@tpetricek What do you think?

The discussion is about using 

```
/// [omit]
```

in case of `markDownComments = false`, where we ignore other markdown constructs.

ademar:
@matthid Sorry, I now realize I did not understood correctly the purpose of `findCommand` and `[omit]`. Let me correct that.

Issue #328, opened 2015-07-07
FSharp.Formatting for F# 3.1 implicitly reading FSharp.Core 4.4.0.0 


dsyme:
I'm using the latest FSharp.Formatting on a Windows machine with both F# 4.0 and F# 3.1 installed. I'm running from a F# 3.1 command prompt.

I'm getting an error where the FSharp.Core 4.4.0.0 (from F# 4.0) is now being used by the FSharp.Formatting toolchain (for example the error below).  This is a bug - the FSharp.Formatting toolchain is currently built on FSharp.Compiler.Service for F# 3.1 and should never be reading FSharp.Core 4.4.0.0 unless explicitly directed to do so.

Has the change to Yaaf.FSharp.Scripting caused this?  Or is it an implicit reference to FSharp.Core 4.4.0.0 flowing from elsewhere, e.g. FCS itself?  Either way the FSharp.Formatting tooling feels somewhat broken right now, or at least not working in an isolated way,

I feel we need to move to a model where the reference set being used by interactive evaluation with FSharp.Formatting is more locked down.

        unknown(1,1): error FS0229: Error opening binary file 'C:\Program Files (x86)\Reference Assemblies\Microsoft\FSharp\.NETFramework\v4.0\4.4.0.0\FSharp.Co re.dll': Error reading/writing metadata for the F# compiled DLL 'C:\Program Files (x86)\Reference Assemblies\Microsoft\FSharp\.NETFramework\v4.0\4.4.0.0\FSharp.Core.dll'. ...).

cheers
don

tpetricek:
My impression was that (when invoked via F# Interactive and not as a stand-alone application) F# Formatting will always load the `FSharp.Core.dll` that is used by the current process. Or is that not the case? 

I think this would be sensible behavior - otherwise, the values you get from `FsiEvaluator` might involve different `FSharp.Core.dll` types (say, list) than the ones your current process is using... (but I'm not sure if this is actually possible?)

matthid:
How can I reproduce this? I have installed both F# 4.0 and F# 3.1 as well, but everything seems to be working fine for me.

I agree with @tpetricek and I tested with Yaaf.FSharp.Scripting with the F# 3.1 fsi.exe and the F# 4.0 fsi.exe and it was always using the current FSharp.Core. But it is quite possible that I got something wrong in the `FSharp.Core` resolution of Yaaf.FSharp.Scripting.

7sharp9:
[...]

Issue #325, opened 2015-06-30
Exception of type 'Microsoft.FSharp.Compiler.Tast+InternalUndefinedItemRef' when generating reference docs for an assembly that uses the Json type provider.


talbottmike:
I'm using FSharp.Formatting as implemented in the ProjectScaffold to generate reference documentation. When attempting to generate reference documentation, I'm getting an exception when MetadataFormat.Generate is called. By process of elimination I found that removing all JsonProviders from my assembly resolves the issue. However, when I re-include a JsonProvider in my project the error returns. I'm not sure that this is an issue with FSharp.Formatting, but any help is appreciated. Thanks

Here's the function from the scaffold generate.fsx where the error occurs.

```
// Build API reference from XML comments
let buildReference () =
  CleanDir (output @@ "reference")
  MetadataFormat.Generate
    ( binaries, output @@ "reference", layoutRootsAll.["en"],
      parameters = ("root", root)::info,
      sourceRepo = githubLink @@ "tree/master",
      sourceFolder = __SOURCE_DIRECTORY__ @@ ".." @@ "..",
      publicOnly = true,libDirs = libDirs )
```

Here's the exception.

```
Microsoft.FSharp.Compiler.ErrorLogger+ReportedError: Exception of type 'Microsoft.FSharp.Compiler.Tast+InternalUndefinedItemRef' was thrown.
   at Microsoft.FSharp.Compiler.ErrorLogger.ErrorLoggerExtensions.ErrorLogger.Error[b](ErrorLogger x, Exception exn) in C:\GitHub\fsharp\FSharp.Compiler.Service\src\fsharp\ErrorLogger.fs:line 321
   at Microsoft.FSharp.Compiler.Tast.EntityRef.get_Deref() in C:\GitHub\fsharp\FSharp.Compiler.Service\src\fsharp\tast.fs:line 2585
   at Microsoft.FSharp.Compiler.Tastops.stripTyEqnsA(TcGlobals g, Boolean canShortcut, TType ty) in C:\GitHub\fsharp\FSharp.Compiler.Service\src\fsharp\tastops.fs:line 614
   at Microsoft.FSharp.Compiler.Tastops.extensionInfoOfTy(TcGlobals g, TType ty) in C:\GitHub\fsharp\FSharp.Compiler.Service\src\fsharp\tastops.fs:line 1441
[...]

PR #332, opened 2015-07-25
Improve logging of FSharp.Formatting

matthid:
With this I plan to fix the mentioned lack of logging (in #325).
I removed the shared `Log.fs` files and added a new `FSharp.Formatting.Common` project to have a single point of configuration for logging (from the outside, as otherwise we have multiple classes at runtime, one in each project).

Another nice sideeffect of this patch is that all the Fharp.Formatting libraries will no longer write into the standard output stream by default (When not explicitly configured with a ConsoleTraceListener). 

Finally I changed the type of some messages to no longer appear in the console (The warning when a documentation for a member could not be found). Those can now be found in the `FSharp.Formatting.svclog` file.

We should start a discussion of what logfiles we want to write and which messages should be printed by default. My current configuration (Setup via `FSharp.Formatting.fsx`) is like this:
- Console: >= Information (Just the message)
- Svclog: ALL (+ALL) -> Will become quite big (several MB), but contains every bit of information
- Logfile: >= Warning (+DateTime)
And all files will be deleted on startup (so only the logfile of the last run is available).
Now thinking about it, the warning-only logfile is probably not that useful.

IMHO this is a nice way to cleanup the output a bit and at the same time have a way to improve debugging.

_TODOs:_
- [x] Switch FSharp.Formatting to `TraceSource`
- [x] Switch Yaaf.FSharp.Scripting to `TraceSource` 
- [x] Add Logging as suggested by #325 (should simplify debugging of #328 as well)

Edit:
It seems like mono doesn't support filtering on TraceSources, this means on mono all messages will be printed to the console, that's a bit unfortunate but not a real problem IMHO.
http://lists.ximian.com/pipermail/mono-list/2012-November/049106.html
http://stackoverflow.com/questions/22590799/mono-tracesource-switch-behaviour


matthid:
@dsyme Regarding your [comment](https://github.com/tpetricek/FSharp.Formatting/issues/325#issuecomment-120325573)

> It uses a secondary in-process F# interactive instance to evaluate stuff during documentation generation, in the context of an existing fsi.exe F# interactive execution of the generate.fsx script. There's definitely potential for problems or gremlins here.

It's actually not that bad, because we can access the complete "fsi.exe" output, so when something fails this output tells us exactly what has happened.

The problem is that we don't have access to the same amount of information via "FSharpChecker". We often have the problem that it works intially, but as soon as we dig into the references (which are the assemblies we want to analyse) it crashes with some internal compiler error nobody is able to decrypt. Is there a trick to access something like the "fsi" stdout?

>  it would be very helpful if FSharp.Formatting printed out the exact command line option set being passed to create the FSharpChecker and also to FsiEvaluationSession.

All this information is now available with this path in a logfile called "FSharp.Formatting.svclog"
 - FSharpChecker: Watch out for entries called "Checker Arguments: "
 - FsiEvaluationSession: Watch out for entries called "Starting nested fsi.exe with args: "

tpetricek:
Very nice work, thank you @matthid! This is going to be super useful for the future of the project (I added logging to R provider and it really helped tremendously debugging things!)

I was a bit worried about adding another `dll`, but looking at your changes - I agree that this was the right thing to do here. I'm just releasing a new version with this, so we'll see what happens when people start using it :-). Hopefully they reference it using the `fsx` file these days!

matthid:
> Hopefully they reference it using the fsx file these days!

Exactly, I couldn't figure out a good way to set the defaults from within the code (without rendering app.config unusable).

Another note is that the logfile is really only useful when opened on windows :). I don't know how good the XmlTraceListener is on the latest mono versions, but in the past I used my own implementation to get a proper logfile on mono (https://github.com/matthid/Yaaf.Logging/blob/develop/src/source/Yaaf.Logging/SimpleXmlWriterTraceListener.fs#L56). So, if we get unusable logfiles we can either switch to raw text or use my implementation.

tpetricek:
Sounds good! In any case, getting a good log in 50% of cases is better than in 0% of cases :-)

Issue #342, opened 2015-10-01
Markdown parser doesn't recognize CodeBlock inside QuotedBlock


ljw1004:
Look at this piece of markdown, and how it renders, and then how it parses in FSharp.Markdown:

```
> __Annotation__
> This is an annnotation
> ```vb
> Module Module1
> End Module
> ```
> More annotation
```

-----------------------------

> __Annotation__
> This is an annnotation
> ```vb
> Module Module1
> End Module
> ```
> More annotation

-------------------------

But FSharp.Markdown parses the quoted block like this:

*   QuotedBlock
    *   Paragraph
        *   Strong
            * Literal "Annotation"
        *   Literal "This is an annotation”
        *   InlineCode "`vb\nModule Module1\nEnd Module"
    *   Literal "More annotation"

It should parse it as a CodeBlock, but it's actually parsing it as an InlineCode.

ljw1004:
It's even worse that that! If the QuotedBlock contains any InlineCode after the CodeBlock, then it parses wrong -- e.g.

```
> Hello
> ```vb
> Module1
> ```
> That was an `inline` code block
```

Here it renders `That was an` as an inline code block, and `inline` as regular text.

ljw1004:
It also fails to parse an indented code block inside a quoted block, for example

```
> Hello
>     Module Module1
>     End Module
> World
```

What I expect here is a CodeBlock inside the QuotedBlock. But instead it rolls the indented code block just into the same single Literal that contains the rest of the stuff.

matthid:
Any PR which brings us closer to the spec (http://spec.commonmark.org/0.22/) is very welcome. Adding unit tests is a big plus :)

tpetricek:
I think that our handling of indented code blocks is correct. You are always required to have a blank line around the code. The CommonMark demo requires this too: [here is a live demo](http://spec.commonmark.org/dingus/?text=%3E%20foo%0A%3E%20%20%20%20%20bar%0A%3E%20zoo%0A%0Aand%20%0A%0A%3E%20foo%0A%3E%20%60%60%60%0A%3E%20bar%0A%3E%20%60%60%60%0A%3E%20zoo).

However, we are not handling the fenced code blocks correctly. We require newlines there too. The following works:

    > Hello
    >
    > ```vb
    > Module1
    > ```
    >
    > That was an `inline` code block

[...]

PR #343, opened 2015-10-01
Follow CommonMark spec.

matthid:
Following Issues like #342 this would be a way to follow the spec.

With some tricks and a .tt file this PR adds code which parses the CommonMark spec html file (http://spec.commonmark.org/0.22/) and creates a test case for each example.

Usually I don't publish my work this early but I hope somebody is willing to help here (currently about 100 of the 600 new tests are passing).

I think this will close a lot of the open issues we have regarding markdown parsing.

@tpetricek Do you like the idea of keeping the generation code or should we only keep the tests?
I think it might be a good idea to be able to update to newer versions of the spec if needed.


tpetricek:
Nice! I do like the idea of automatically generating the tests from the specification so that we can handle potential changes. Too bad that the specification is not in a pretty machine readable format....

I have not used T4 before - does it work on mono too? An alternative would be to use `TestCaseSource` like in the [existing test that compares PHP Markdown parser files](https://github.com/tpetricek/FSharp.Formatting/blob/master/tests/FSharp.Markdown.Tests/Externals.fs). But I guess both will work here.

matthid:
`TestCaseSource` is definitely the way to go, thank you for the hint. That's waaaay better :)

matthid:
Changed the test generation again and now use the official approach to run python to generate a test-validation json file (https://github.com/jgm/CommonMark#running-tests-against-the-spec).
The json file is consumed via the FSharp.Data json provider.
This simplifies the code a bit and gives some more metadata (like section of the spec for the given test).

I have changed FSF in a way to follow the spec for the 'Fenced code blocks' and 'Indented code blocks' sections (all other sections are disabled for now, see https://github.com/tpetricek/FSharp.Formatting/blob/master/tests/FSharp.Markdown.Tests/CommonMarkSpecTest.fs#L20)

It's probably better to merge this now and handle section by section in the future (we may even get contributions for other sections by other people).

PR #345, opened 2015-10-05
Allow all 1.3.x versions of FCS

forki:


matthid:
LGTM, ideally/eventually FSharpVSPowerTools.Core will release a F#4 compatible version so we can get back to latest.

tpetricek:
This looks reasonable!

PR #372, opened 2015-12-29
Add ComiledName to members with F# specific naming

artyomszasa:
F# allows to specify CompiledName for variables/members/etc..
It would be usefull to display the CompiledName -- it is important when accessing the entry from a .NET language other than F# (msdn quote)...

MSDN does so, e.g.  https://msdn.microsoft.com/en-us/library/ee370346.aspx
> This function is named Map in compiled assemblies. If you are accessing the function from a .NET language other than F#, or through reflection, use this name.

Note:
The compiled names are only added when the CompiledName differs from the DisplayName.

### DIFF
--- a/misc/templates/reference/part-members.cshtml
+++ b/misc/templates/reference/part-members.cshtml
@@ -33,9 +33,9 @@
           }
           @it.Comment.FullText
           @if (!String.IsNullOrEmpty(it.Details.FormatCompiledName))
-		  {
-			  @:<p>CompiledName: <code>@it.Details.FormatCompiledName</code></p>
-		  }
+          {
+              @:<p>CompiledName: <code>@it.Details.FormatCompiledName</code></p>
+          }
         </td>
       </tr>
     }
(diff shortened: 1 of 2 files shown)


misc/templates/reference/part-members.cshtml
```diff
@@ -32,9 +32,13 @@
             </a>
           }
           @it.Comment.FullText
+          @if (!String.IsNullOrEmpty(it.Details.FormatCompiledName))
+		  {
+			  @:<p>CompiledName: <code>@it.Details.FormatCompiledName</code></p>
+		  }
```
matthid: looks like broken whitespace

matthid:
Sure why not. I like it.

matthid:
Do you need a fast release with this?

artyomszasa:
Sure it would be nice to see the changes in the release soon. But I don't really *need* it -- our doc generation setup script can work with both nuget and custom built assemblies.

matthid:
Released in https://www.nuget.org/packages/FSharp.Formatting/2.13.0

PR #377, opened 2016-01-12
Make logging to file optional using environment variable

tpetricek:
Following the discussion here: https://github.com/tpetricek/FsSnip.Website/issues/32

I would prefer to keep the logging to a file optional. I added a new documentation page that explains how to enable this so that when people get an error, we can point them to the docs on how to get a detailed log :)


tpetricek:
@matthid I know you would have preferred the logging to be on by default - this was causing me some troubles in deploying a new version of fssnip.net to Azure, so I decided to go this way. I think it is a sensible default, but the idea with adding two load scripts would work too, so feel free to do that if you think it is worth it.

The doc page is now live: http://tpetricek.github.io/FSharp.Formatting/diagnostics.html

matthid:
Yes I prefered an opt-out but I don't think it's important enough to have two load scripts. The docs are really clear so I'll just enable the flag in my builds :)

PR #381, opened 2016-01-25
Fixes issues in PaketFormat

LimoWanKenobi:
Fixes issues reported in fsprojects/Paket#1419

 - Adds "HTTP" to list of Paket keywords.
 - Adds :/ to list of prefixes to check that are not present when matching a comment
 - Extends test of Paket coloring.


tpetricek:
Thanks! A new release 2.13.5 with this included should be on NuGet in a minute or two!

PR #382, opened 2016-02-01
add test for #380 and change the fix.

matthid:


LimoWanKenobi:
:disappointed: I completely missed this case!

matthid:
@LimoWanKenobi and the `Test` attribute :P , but no worries :)

PR #386, opened 2016-02-29
Added TypeScript to the CSharpFormat project

mktange:
Added TypeScriptFormat since it is needed for one of the projects I am working on.

Example of it being used can be seen here:
http://delegateas.github.io/Delegate.XrmDefinitelyTyped/xrm-dts-usage.html


tpetricek:
Fantastic! 

Thanks a lot for adding this. I just released this as version 2.13.6.

PR #394, opened 2016-04-03
add code to parse table rows correctly

tomcl:
Small change to table row parsing function replacing split on delimiter
by parse for escapes and inlines

kludged test directory added under tests - this is not done properly and could be deleted but would need some new tests of complex tables!


src/FSharp.Markdown/MarkdownParser.fs
```diff
@@ -449,6 +449,35 @@ let rec (|ListItems|_|) prevSimple = function
 
 // Code for parsing pipe tables
 
+// Splits table row into deliminated parts escaping inline code and latex
+let rec PipeTableFindSplits (delim : char array) (line : char list) =
```
matthid: Can we lower case this (or change to an active pattern)?

matthid:
Thanks for the patch!
Can we somehow integrate your tests into the existing test suite?

src/FSharp.Markdown/MarkdownParser.fs
tomcl: Matthias,

I made a "minimal change" patch so this is a subfunction used by the 
existing row parser - it is not a recogniser.

It could be inserted as a subfunction of active recogniser PipeTableRow, 
but big subfunctions inside smaller functions are not so readable?

Best wishes, Tom

matthid: Its more about the code-style: The function is fine but please call it "pipeTableFindSplits" instead of "PipeTableFindSplits". Only active pattern are a exception here.

tomcl: No problem. Do I create a new pull request or what?

matthid: You can just commit another commit and push it to the same branch you started the pull request with.

tpetricek:
Just to make sure I understand this - is this allowing parsing of tables that were not parsed before?

I know someone just emailed me that the following does not parse as table:

    | Alpha   | Beta |
    |---------|------|
    | `x & y` | And  |
    | `x | y` | Or   |

... but I guess this is unrelated?

tomcl:
On 05/04/2016 14:47, Tomas Petricek wrote:
>
> Just to make sure I understand this - is this allowing parsing of 
> tables that were not parsed before?
>

Yes - that is correct.

The existing table row parser does not use inline parsing (with inline 
code,math) to parse table rows. I've rewritten the part of this that 
detects table delimiters so that when these are inside inline code or 
escaped they do not form new columns.
>
> I know someone just emailed me that the following does not parse as table:
>
> || Alpha | Beta | |---------|------| | `x & y` | And | | `x | y` | Or | |
>
> ... but I guess this is unrelated?
>
That is a case I deal with.

tomcl:
So - is there any reason now why this should not be merged?

tpetricek:
No, just my slowness after getting back from holidays :-).

It will be on NuGet as 2.14.2 in a minute. Thanks a lot for the contribution!

PR #395, opened 2016-04-05
Fix FCS dependency

forki:


tpetricek:
Does anybody have any idea why the new FCS release caused troubles? cc @matthid @dsyme 

Even though it is only 2.0.0.6 -> 2.0.0.8 version change, there seem to be some API changes, like rename of `Dependencies` to `AllDependenciesDeprecated` in `IncrementalBuild.fsi` and the change of type of `GetLogicalTimeStamp`. See [the diff here](https://github.com/fsharp/FSharp.Compiler.Service/compare/2.0.0.6...master#diff-8dbcf18e6248552a193c320a007b453b).

F# Formatting is not using those, but perhaps PowerTools depend on some of them? (cc @dungpa @vasily-kirichenko), in which case we'd need a new release of VFPT...

vasily-kirichenko:
We updated FCS in Fantomas, FSharpLint and VFPT to 2.0.0.8 a couple of days ago. 
@dungpa is it possible to publish a new VFPT.Core Nuget package without publishing new VFPT itself?

tpetricek:
@vasily-kirichenko @dungpa Thanks! That would be amazing.

(Though I'm not 100% sure this is an issue - it might be something on our side too!)

dungpa:
@vasily-kirichenko Yes, you could build the project in Release mode, comment out this line https://github.com/fsprojects/VisualFSharpPowerTools/blob/e3b2531d83e008360c1fd6bcdb0bcfb25c9aaa31/build.fsx#L337 and then run PublishNuGet target.

vasily-kirichenko:
@dungpa Thanks, done https://www.nuget.org/packages/FSharpVSPowerTools.Core/2.4.0

@tpetricek try to update the package, I hope it will help.

Issue #355, opened 2015-11-12
Markdown.TransformHtml() unexpectedly stopped


cnblogs-dudu:
Unable to transform the markdown text in the following C# code  to html.
```csharp
namespace CNBlogs.Markdown.Tests
{
    public class FSharpMarkdownTest
    {
        [Fact]
        public void TransformHtmlTest()
        {
            var markdownText = @"- controller

 > demo
 > test。
>
>，";
            var html = FSharp.Markdown.Markdown.TransformHtml(markdownText);
            Console.WriteLine(html);
        }
    }
}
```
Get the following error:
```
Test 'CNBlogs.Markdown.Tests.FSharpMarkdownTest.TransformHtmlTest' failed: System.Exception : Unexpectedly stopped!
[""; "，"]
	at FSharp.Markdown.Parser.parseParagraphs@639-12.Invoke(String message)
	at Microsoft.FSharp.Core.PrintfImpl.go@523-3[b,c,d](String fmt, Int32 len, FSharpFunc`2 outputChar, FSharpFunc`2 outa, b os, FSharpFunc`2 finalize, FSharpList`1 args, Int32 i)
	at Microsoft.FSharp.Core.PrintfImpl.run@521[b,c,d](FSharpFunc`2 initialize, String fmt, Int32 len, FSharpList`1 args)
	at <StartupCode$FSharp-Core>.$Reflect.Invoke@720-4.Invoke(T1 inp)
	at FSharp.Markdown.Parser.parseParagraphs@568.Invoke(Unit unitVar)
	at Microsoft.FSharp.Collections.SeqModule.mkDelayedSeq@835.Invoke(Unit unitVar0)
	at Microsoft.FSharp.Collections.SeqModule.ToList[T](IEnumerable`1 source)
	at FSharp.Markdown.Parser.parseParagraphs@568.Invoke(Unit unitVar)
	at Microsoft.FSharp.Collections.SeqModule.mkDelayedSeq@835.Invoke(Unit unitVar0)
	at Microsoft.FSharp.Collections.SeqModule.Append@1061-1.Invoke(Unit unitVar0)
	at Microsoft.FSharp.Collections.Generator.Microsoft-FSharp-Collections-Generator-Generator-1-get_Apply@386.Invoke(Unit unitVar0)
	at Microsoft.FSharp.Collections.Generator.appG[a](Generator`1 g)
[...]

PR #400, opened 2016-04-23
End blockquote on a line with an empty blockquote. For #355

theprash:
This stops an exception being thrown but it also changes the behaviour for this case:

```fsharp
">a
>
a"
```
Previous output:
```html
<blockquote>
<p>a</p>
<p>a</p>
</blockquote>
```
New output:
```html
<blockquote>
<p>a</p>
</blockquote>
<p>a</p>
```

I'm not sure if we want this behaviour change as it's a bit of a weird edge case, but it does agree with: http://spec.commonmark.org/dingus/

### DIFF
--- a/src/FSharp.Markdown/MarkdownParser.fs
+++ b/src/FSharp.Markdown/MarkdownParser.fs
@@ -630,12 +630,21 @@ let (|LinesUntilBlockquoteEnds|) input =
 /// Recognizes blockquote - continues taking paragraphs
 /// starting with '>' until there is something else
 let rec (|Blockquote|_|) = function
+  | EmptyBlockquote(Lines.TrimBlankStart rest) ->
+      Some ([""], rest)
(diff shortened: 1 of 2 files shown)


theprash:
I noticed there are disabled CommonMark spec tests. It seems like the long term goal is to conform to this.

I tried enabling the tests and found that my change caused one more test case to pass (line 1595 of `commonmark_spec.json`), where the input was:

```
> bar
>
baz

```

matthid:
Yes the long term goal is to conform to the spec, so IMHO this is a good patch. Thanks.
Sadly travis seems to be failing right now.

matthid:
Ok restarted Travis, everything is green :)

matthid:
Do you need a release with this or do you have workarounds already in place?

theprash:
Achievement unlocked: first open source contribution. Thanks :)

The fix wasn't for me. I just picked up the issue logged 5 months ago so I wouldn't worry about a release.

PR #404, opened 2016-05-14
Enabled and fixed 3 CommonMark test sections

theprash:
I enabled "Paragraphs", "Precedence" and "Tabs" and fixed some failing tests.

Note that while fixing `(|BlockquoteStart|_|)`, I rewrote it with regex as it seemed much clearer.

### DIFF
--- a/FSharp.Formatting.sln
+++ b/FSharp.Formatting.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.25123.0
 MinimumVisualStudioVersion = 12.0.31101.0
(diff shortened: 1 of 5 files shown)


tpetricek:
Sorry for the delay here - this looks amazing!

I guess the one thing that's now different is that `BlockquoteStart` will also match code that is actually inline code (4 spaces followed by `>`). This does not seem to be an issue (we probably look for code before looking for blockquote), but it might be safer to allow only up to 3 spaces in the ` *` part of the regex. Or does the CommonMark spec allow more spaces?

theprash:
There's no harm in limiting it to three leading spaces, so I've made that change. I also added a test for a blockquote with leading spaces.

matthid:
Wow very nice! A good step forward in following the CommonMark spec :+1: 
The changes look good to me :shipit:

tpetricek:
Fantastic! Thank you very much for the awesome PR!

theprash:
You're welcome. Thanks to you both for reviewing and accepting it :smiley:

PR #406, opened 2016-05-15
Fix tooltip flickering in CSS

### LINKED ISSUE #405: Ugly flicker when tooltip overlaps parent element
When FSharp.Formatting is used to generate HTML markup, its tooltips look fine if they don't overlap their parent element. But if, due to their positioning on the page or the CSS used to style them, they end up overlapping their parent element, some ugly flicker is visible as the mouse pointer moves over the parent element. Here's a visual example from [FsReveal](https://github.com/fsprojects/FsReveal), which uses FSharp.Formatting to produce its slides:

bad

That's from FsReveal's sample presentation, with no changes made. The mouse cursor failed to show up in this animated screen capture, but I was moving it back and forth between the name `factorial` and the `let` just to the left.

This problem is also visible in the FSharp.Formatting documentation, although it's harder to see it happening there because the tooltips *almost* don't overlap their parent elements. But go to http://tpetricek.github.io/FSharp.Formatting/literate.html and hover the mouse over any F# name that has a tooltip (say, `printfn` in the first code sample). The tooltip (at least on my browser) overlaps the parent `span` by one row of pixels, making it possible to trigger the flickering by moving the mouse very precisely left and right alone that one row of pixels. Like so:

literate

The cause of this issue is that when the tooltip element appears, it naturally lies above its parent element in z-index order and thus the browser considers it to be what the mouse cursor is currently pointing to. When the element appears, as long as the cursor hasn't moved yet, no `mouseover` or `mouseout` events will fire. But the moment the cursor moves even one pixel, the browser sees it as over a new element: it *was* over the parent `span`, and now it's over the tooltip `div`. So the browser fires a `mouseout` event on the parent `span` (which causes the tooltip to be hidden) and a `mouseover` event on the tooltip `div` (which does nothing). And now the tooltip is hidden, so the next time the cursor moves, a `mouseout` will fire on the tooltip `div` (doing nothing) and a `mouseover` will fire on the parent `span` (showing the tooltip again). And that cycle repeats all over again, causing the tooltip `div` to alternate between shown and hidden with every update of the mouse pointer. Result: highly-annoying flickering.

There are two possible solutions to this issue. One is to take advantage of the little-known `pointer-events` CSS property. An element with `pointer-events: none` will be ignored for the purpose of `mouseover` and `mouseout` events (and other mouse events as well). This is the simplest solution since it only involves adding a single line of CSS to the `.tip` class, and it works on modern versions of all browsers. It does **not**, however, work on IE versions 10 or earlier; only IE 11+ will honor `pointer-events: none` properly on HTML elements, according to the http://caniuse.com/#feat=pointer-events chart.

That solution produces the following result:

good

It's not quite obvious in that animated GIF, but the tooltip disappears as soon as the mouse moves off the word `factorial`. It would be more obvious if I had moved the mouse more slowly, but the tooltip can disappear while the mouse cursor is still over the tooltip, as long as the cursor has moved off the parent element (in that example, the `span` containing the word `factorial`).

Another possible solution would be to change the logic of the tooltips in Javascript: instead of disappearing when the mouse leaves the parent element, make the tooltip stay visible as long as the mouse is inside the parent element *or* inside the tooltip. In other words, the tooltip would behave like this:

also-good

Again, it may not be obvious, but here the tooltip stays visible even when the mouse cursor has left the parent `span`.

The drawback of this approach is that if the code in question has a lot of invocations of common functions like `List.reduce` or `Seq.ofList`, the tooltips for `List` and `Seq` will tend to be quite large -- and it's possible that someone trying to hover the mouse pointer over `reduce` will accidentally brush it over the name `List` first, and then have to move the pointer quite far away before they get off the tooltip. At which point getting back to the word `reduce` can become an exercise in frustration, as any module name becomes a landmine the mouse pointer has to avoid.

Therefore, I would suggest the CSS fix (adding `pointer-events: none` to the tooltip CSS). Its only drawback is that it won't work in IE versions before 11. But if anyone is running IE 10 or earlier, they really should upgrade anyway, so I don't find that to be a major issue. Therefore, my recommendation would be to use the CSS fix rather than the Javascript fix.

rmunn:
Fixes #405.

This solution uses `pointer-events: none` to ensure that tooltips don't receive mouseover and mouseout events, thereby removing the flicker that occurs when the tooltip overlaps its parent element.

Browser compatibility: http://caniuse.com/#feat=pointer-events

### DIFF
--- a/docs/files/content/style.css
+++ b/docs/files/content/style.css
@@ -54,6 +54,7 @@ div.tip {
 	padding:6px 8px 6px 8px;
 	display:none;
   color:#d1d1d1;
+  pointer-events:none;
 }
 table.pre pre {
   padding:0px;
--- a/docs/files/content/style_light.css
+++ b/docs/files/content/style_light.css
@@ -55,6 +55,7 @@ div.tip {
 	padding:6px 8px 6px 8px;
 	display:none;
   color:#000000;
+  pointer-events:none;
 }
 table.pre pre {
   padding:0px;
--- a/src/FSharp.CodeFormat/files/style.css
+++ b/src/FSharp.CodeFormat/files/style.css
@@ -36,6 +36,7 @@ div.tip
 	border:1px solid #606060;
 	background:#ffffd0;
 	display:none;
+	pointer-events:none;
 }
 
 h2


rmunn:
Note that this PR and PR #407 are mutually-exclusive: if you set `pointer-events: none` on the `.tip` class, it won't be able to handle any mouse events, and PR #407 needs the tooltip `div` to handle mouse events.

PR #411, opened 2016-06-19
Add range to MarkdownParagraph and MarkdownSpan

AviAvni:
this is adding only the line to the range please review before adding more information
issue: https://github.com/tpetricek/FSharp.Formatting/issues/410
@tpetricek

### DIFF
--- a/docs/content/codeformat.fsx
+++ b/docs/content/codeformat.fsx
@@ -24,7 +24,7 @@ can be called to format snippets repeatedly:
 *)
 
 let fsharpCompiler = Assembly.Load("FSharp.Compiler")
-let formattingAgent = CodeFormat.CreateAgent(fsharpCompiler)
+let formattingAgent = CodeFormat.CreateAgent()
 
 (**
 If you want to process multiple snippets, it is a good idea to keep the 
--- a/docs/content/markdown.fsx
+++ b/docs/content/markdown.fsx
@@ -10,7 +10,10 @@ First, we need to load the assembly and open necessary namespaces:
 *)
 
 #r "../../bin/FSharp.Markdown.dll"
+#r "../../bin/FSharp.Formatting.Common.dll"
 open FSharp.Markdown
+open FSharp.Formatting.Common
+
 
 (**
 Parsing documents
@@ -64,7 +67,7 @@ The following snippet prints the heading of the document:
 // Iterate over all the paragraph elements
 for par in parsed.Paragraphs do
   match par with
-  | Heading(1, [Literal text]) -> 
+  | Heading(1, [Literal(text, _)], _) -> 
       // Recognize heading that has a simple content
       // containing just a literal (no other formatting)
       printfn "%s" text
@@ -92,8 +95,8 @@ to recognize any paragraph or span that can contain child elements:
 /// Returns all links in a specified span node
 let rec collectSpanLinks span = seq {
   match span with
-  | DirectLink(_, (url, _)) -> yield url
-  | IndirectLink(_, _, key) -> yield fst (parsed.DefinedLinks.[key])
+  | DirectLink(_, (url, _), _) -> yield url
+  | IndirectLink(_, _, key, _) -> yield fst (parsed.DefinedLinks.[key])
   | Matching.SpanLeaf _ -> ()
   | Matching.SpanNode(_, spans) ->
       for s in spans do yield! collectSpanLinks s }
--- a/docs/content/markdown.fsx
+++ b/docs/content/markdown.fsx
@@ -67,7 +67,7 @@ The following snippet prints the heading of the document:
 // Iterate over all the paragraph elements
 for par in parsed.Paragraphs do
   match par with
-  | Heading(1, [Literal(text, _)], _) -> 
+  | Heading(size=1; body=[Literal(text=text)]) -> 
       // Recognize heading that has a simple content
       // containing just a literal (no other formatting)
       printfn "%s" text
--- a/src/Common/StringParsing.fs
+++ b/src/Common/StringParsing.fs
@@ -14,9 +14,83 @@ open FSharp.Collections
 
 module String =
   /// Matches when a string is a whitespace or null
-  let (|WhiteSpaceS|_|) (s) = 
+  let (|WhiteSpace|_|) (s) = 
     if String.IsNullOrWhiteSpace(s) then Some() else None
 
+  /// Returns a string trimmed from both start and end
+  let (|TrimBoth|) (text:string) = text.Trim()
+
+  /// Matches when a string starts with the specified sub-string
+  let (|StartsWith|_|) (start:string) (text:string) = 
+    if text.StartsWith(start) then Some(text.Substring(start.Length)) else None
+
+  /// Matches when a string starts with the specified sub-string
+  /// The matched string is trimmed from all whitespace.
+  let (|StartsWithTrim|_|) (start:string) (text:string) = 
+    if text.StartsWith(start) then Some(text.Substring(start.Length).Trim()) else None
+
+  /// Matches when a string starts with the given value and ends 
+  /// with a given value (and returns the rest of it)
+  let (|StartsAndEndsWith|_|) (starts, ends) (s:string) =
+    if s.StartsWith(starts) && s.EndsWith(ends) && 
+       s.Length >= starts.Length + ends.Length then 
+      Some(s.Substring(starts.Length, s.Length - starts.Length - ends.Length))
+    else None
+
+  /// Matches when a string starts with the given value and ends 
+  /// with a given value (and returns trimmed body)
+  let (|StartsAndEndsWithTrim|_|) args = function
+    | StartsAndEndsWith args (TrimBoth res) -> Some res
+    | _ -> None
+
+  /// Matches when a string starts with a sub-string wrapped using the 
+  /// opening and closing sub-string specified in the parameter.
+  /// For example "[aa]bc" is wrapped in [ and ] pair. Returns the wrapped
+  /// text together with the rest.
+  let (|StartsWithWrapped|_|) (starts:string, ends:string) (text:string) = 
+    if text.StartsWith(starts) then 
+      let id = text.IndexOf(ends, starts.Length)
+      if id >= 0 then 
+        let wrapped = text.Substring(starts.Length, id - starts.Length)
+        let rest = text.Substring(id + ends.Length, text.Length - id - ends.Length)
+        Some(wrapped, rest)
+      else None
+    else None
+
+  /// Ignores everything until a end-line character is detected, returns the remaining string.
+  let (|SkipSingleLine|) (text:string) =
+    let rec tryEol eolList =
+      match eolList with
+      | h : string :: t ->
+        match text.IndexOf(h) with
+        | i when i < 0 -> tryEol t
+        | i ->
+          text.Substring (i + h.Length)
+      | _ ->
+        text
+    let result = tryEol [ "\r\n"; "\n" ]
+    let skipped = text.Substring(0, text.Length - result.Length)
+    if not <| String.IsNullOrWhiteSpace(skipped) then
+      FSharp.Formatting.Common.Log.warnf "skipped '%s' which contains non-whitespace character!" skipped
+    if result = text then
+      FSharp.Formatting.Common.Log.warnf "could not skip a line of %s, because no line-ending character was found!" text
+    result
+
+  /// Given a list of lines indented with certan number of whitespace 
+  /// characters (spaces), remove the spaces from the beginning of each line 
+  /// and return the string as a list of lines
+  let removeSpaces lines =
+    let spaces =
+      lines 
+      |> Seq.filter (String.IsNullOrWhiteSpace >> not)
+      |> Seq.map (fun line -> line |> Seq.takeWhile Char.IsWhiteSpace |> Seq.length)
+      |> fun xs -> if Seq.isEmpty xs then 0 else Seq.min xs
+    lines 
+    |> Seq.map (fun line -> 
+        if String.IsNullOrWhiteSpace(line) then ""
+        else line.Substring(spaces))
+
+module StringPosition =
   /// Matches when a string is a whitespace or null
   let (|WhiteSpace|_|) (s, n: int) = 
     if String.IsNullOrWhiteSpace(s) then Some() else None
@@ -24,9 +98,7 @@ module String =
   /// Matches when a string does starts with non-whitespace
   let (|Unindented|_|) (s:string, n:int) = 
     if not (String.IsNullOrWhiteSpace(s)) && s.TrimStart() = s then Some() else None
-
-  /// Returns a string trimmed from both start and end
-  let (|TrimBothS|) (text:string) = text.Trim()
+      
   /// Returns a string trimmed from both start and end
   let (|TrimBoth|) (text:string, n:int) = (text.Trim(), n)
   /// Returns a string trimmed from the end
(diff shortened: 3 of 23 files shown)


ljw1004:
It looks like you're storing the range exactly how I'd want in the datastructures, and passing around the codebase how I'd expect.

Although: I hope you store range information solely with a character offset into the file. It feels like line/col information is *derived* from that. I suspect that a lot of book-keeping will be easier if the primary data structure is character-offsets. It's also how Roslyn does it :). (There'd probably be an additional datastructure generated during parsing which is the "character offset at which each line in the file starts", for an easy `OffsetToLineAndCol` function).

Presumably you're planning to add full range information next, i.e file-offset of both the start and end of each construct? I guess you'd have to start by abandoning `ReadLines()` and switching to your own `ReadLinesWithRangeInformation()`. And plumbing through ParseParagraphs to store start+end. And altering ParseChars so it knows which character it's on.

It's a change that has lots of tendrils everywhere, isn't it?!

AviAvni:
@ljw1004 thanks about the comment I'm doing it as a thanks to your great job in Roslyn
It was the first time I saw this code so to know more how it's works I start implement it without doing big changes to the original code which read lines only add the needed information
Your suggesting do the parsing differently I'll discuss with @tpetricek what the best way

src/Common/StringParsing.fs
```diff
@@ -14,44 +14,57 @@ open FSharp.Collections
...
-    len, text.Substring(0, len).Replace("\t", "    ").Length, trimmed
+    len, text.Substring(0, len).Replace("\t", "    ").Length, (trimmed, n)
 
   /// Matches when a string starts with any of the specified sub-strings
-  let (|StartsWithAny|_|) (starts:seq<string>) (text:string) = 
+  let (|StartsWithAny|_|) (starts:seq<string>) (text:string, n:int) = 
     if starts |> Seq.exists (text.StartsWith) then Some() else None
   /// Matches when a string starts with the specified sub-string
-  let (|StartsWith|_|) (start:string) (text:string) = 
+  let (|StartsWithS|_|) (start:string) (text:string) =
```
tpetricek: I think the `S` suffix on the active patterns is not very idiomatic. Could we instead have two modules, say `String` and `StringPosition` so that the patterns that work on string with position are all in `StringPosition` module?

src/FSharp.Markdown/Markdown.fs
```diff
@@ -13,6 +13,8 @@ open System.Collections.Generic
 // Definition of the Markdown format
 // --------------------------------------------------------------------------------------
 
+type MarkdownRange = { Line : int }
```
tpetricek: Does `Line` here mean just the line number? As @ljw1004 mentioned, it would make sense to track the total offset (in number of characters) from the beginning of the file.

As Markdown is fairly line-based, I guess we could keep line & column number for the position (which might be easier to implement on top of the current parser). Also, one thing to consider is whether we want just the _start_ of a span or whether we want _start and end_ (I think having both is better idea.)

src/FSharp.Markdown/Markdown.fs
```diff
@@ -28,19 +30,19 @@ type MarkdownColumnAlignment =
...
-  | AnchorLink of string 
-  | DirectLink of MarkdownSpans * (string * option<string>)
-  | IndirectLink of MarkdownSpans * string * string
-  | DirectImage of string * (string * option<string>)
-  | IndirectImage of string * string * string
-  | HardLineBreak
-  | LatexInlineMath of string
-  | LatexDisplayMath of string
-  | EmbedSpans of MarkdownEmbedSpans
+  | Literal of string * MarkdownRange option
```
tpetricek: As this is a large-ish breaking change, it would be great to switch to the F# named parameters on unions, to make it less breaking-change-prone in the future. Could you please add names to the parameters, e.g.:

    | Literal of text:string * range:MarkdownRange option
    | InlineCode of code:string * range:MarkdownRange option
    | Strong of body:MarkdownSpans * range:MarkdownRange option

... etc. I'll be happy to review the naming on the next pass.

tpetricek: (This will be a great improvement in general, because it'll be possible to extract information about nodes based on the names, but this is only available in F# since 2015 (I think?)).

src/FSharp.Markdown/MarkdownParser.fs
```diff
@@ -145,107 +145,113 @@ let (|Emphasised|_|) = function
     | _ -> None
   | _ -> None
 
+/// Defines a context for the main `parseParagraphs` function
+type ParsingContext =
```
tpetricek: +1 This looks like a good way of keeping the context :)

tpetricek:
Sorry for the long delay - I finally had a look at this. 

 - I had one style comment, which is that the active patterns ending with `S` suffix do not fit well with the F# coding style - I propose to instead have `String` and `StringPosition` module that contain two versions of the patterns.

 - I got the impression that the current code tracks just the line numbers. I think that's not optimal - we need either line and column numbers, or offset in characters from the beginning of the file. Also, I think range should have start position and end position.

 - While we're doing this breaking change, it would be nice to add names on the DU cases - this way, future changes won't be breaking :)

ljw1004:
I don't have further technical comments, Avi, but just want to say how awesome it is that you're doing this.

src/Common/StringParsing.fs
AviAvni: Done

src/FSharp.Markdown/Markdown.fs
AviAvni: Done

AviAvni:
@tpetricek @ljw1004 Hi thanks for reviewing the code
I Implement all the issue please review again

docs/content/codeformat.fsx
```diff
@@ -24,7 +24,7 @@ can be called to format snippets repeatedly:
 *)
 
 let fsharpCompiler = Assembly.Load("FSharp.Compiler")
-let formattingAgent = CodeFormat.CreateAgent(fsharpCompiler)
+let formattingAgent = CodeFormat.CreateAgent()
```
tpetricek: Out of curiosity, is this a needed change for some reason? (Happy to delete the parameter if it's not needed, but then we should probably delete the whole line 26.)

src/FSharp.Markdown/Markdown.fs
```diff
@@ -28,19 +29,19 @@ type MarkdownColumnAlignment =
...
-  | DirectImage of string * (string * option<string>)
-  | IndirectImage of string * string * string
-  | HardLineBreak
-  | LatexInlineMath of string
-  | LatexDisplayMath of string
-  | EmbedSpans of MarkdownEmbedSpans
+  | Literal of text:string * range:MarkdownRange option
+  | InlineCode of code:string * range:MarkdownRange option
+  | Strong of spans:MarkdownSpans * range:MarkdownRange option
+  | Emphasis of spans:MarkdownSpans * range:MarkdownRange option
```
tpetricek: Could you use `body` rather than `spans` for `Strong`, `Emphasis`, `DirectLink` and `IndirectLink`?

src/FSharp.Markdown/Markdown.fs
```diff
@@ -28,19 +29,19 @@ type MarkdownColumnAlignment =
...
+  | AnchorLink of link:string * range:MarkdownRange option
+  | DirectLink of spans:MarkdownSpans * linkAndTitle:(string * option<string>) * range:MarkdownRange option
+  | IndirectLink of spans:MarkdownSpans * link:string * key:string * range:MarkdownRange option
+  | DirectImage of body:string * linkAndTitle:(string * option<string>) * range:MarkdownRange option
```
tpetricek: I would probably prefer to split `linkAndTitle` into two: `link:string * title:option<string>`

src/FSharp.Markdown/Markdown.fs
```diff
@@ -54,17 +55,17 @@ and MarkdownEmbedSpans =
...
-  | CodeBlock of string * string * string
-  | InlineBlock of string
-  | ListBlock of MarkdownListKind * list<MarkdownParagraphs>
-  | QuotedBlock of MarkdownParagraphs
-  | Span of MarkdownSpans
-  | LatexBlock of list<string>
-  | HorizontalRule of char
-  | TableBlock of option<MarkdownTableRow> * list<MarkdownColumnAlignment> * list<MarkdownTableRow>
-  | EmbedParagraphs of MarkdownEmbedParagraphs
+  | Heading of n:int * spans:MarkdownSpans * range:MarkdownRange option
```
tpetricek: I think `size` is better name than `n` - and as above, I prefer `body` over `spans` for `Heading`, `Paragraph` and `Span`.

src/FSharp.Markdown/Markdown.fs
```diff
@@ -54,17 +55,17 @@ and MarkdownEmbedSpans =
...
+  | Paragraph of spans:MarkdownSpans * range:MarkdownRange option
+  | CodeBlock of code:string * language:string * ignoredLine:string * range:MarkdownRange option
+  | InlineBlock of code:string * range:MarkdownRange option
+  | ListBlock of kind:MarkdownListKind * items:list<MarkdownParagraphs> * range:MarkdownRange option
+  | QuotedBlock of paragraphs:MarkdownParagraphs * range:MarkdownRange option
+  | Span of spans:MarkdownSpans * range:MarkdownRange option
+  | LatexBlock of body:list<string> * range:MarkdownRange option
+  | HorizontalRule of c:char * range:MarkdownRange option
```
tpetricek: I'd prefer `character` over `c` here.

docs/content/markdown.fsx
```diff
@@ -64,7 +67,7 @@ The following snippet prints the heading of the document:
 // Iterate over all the paragraph elements
 for par in parsed.Paragraphs do
   match par with
-  | Heading(1, [Literal text]) -> 
+  | Heading(1, [Literal(text, _)], _) ->
```
tpetricek: To encourage people writing the code in the new style - and future-proof way, could you change the pattern matching here to use the naming? For example:

    | Heading(size=1; body=[Literal(text=text)]) -> 

I know this is quite tedious (!) but adding the names to the DUs is a fantastic improvement for the library and it would be great to have it in a couple of places in the docs, so that people know about it - and use it!

src/FSharp.Formatting.Common/Range.fs
```diff
@@ -0,0 +1,11 @@
+﻿namespace FSharp.Formatting.Common
+
+type MarkdownRange = { StartLine : int; StartColumn : int; EndLine : int; EndColumn : int }
+
+module MRange =
```
tpetricek: A slightly more idiomatic thing would be to name the module `MarkdownRange` (same as the type name), which requires the following attribute:

    [<CompilationRepresentation (CompilationRepresentationFlags.ModuleSuffix)>]
    module MarkdownRange =
       (...)

src/FSharp.Markdown/HtmlFormatting.fs
```diff
@@ -67,19 +67,19 @@ let noBreak (ctx:FormattingContext) () = ()
...
-  | Literal(str) -> ctx.Writer.Write(str)
-  | HardLineBreak -> ctx.Writer.Write("<br />" + ctx.Newline)
-  | IndirectLink(body, _, LookupKey ctx.Links (link, title)) 
-  | DirectLink(body, (link, title)) -> 
+  | AnchorLink(id, _) -> ctx.Writer.Write("<a name=\"" + id + "\">&#160;</a>") 
+  | EmbedSpans(cmd, _) -> formatSpans ctx (cmd.Render())
+  | Literal(str, _) -> ctx.Writer.Write(str)
+  | HardLineBreak(_) -> ctx.Writer.Write("<br />" + ctx.Newline)
+  | IndirectLink(body, _, LookupKey ctx.Links (link, title), _) 
+  | DirectLink(body, (link, title), _) ->
```
tpetricek: We definitely do _not_ need to use the named syntax here - adding `_` is just fine here :). 

(I think it is important to have the named syntax in the docs, but not everywhere in the codebase.)

src/FSharp.Markdown/MarkdownParser.fs
```diff
@@ -145,107 +146,128 @@ let (|Emphasised|_|) = function
...
 /// Parses a body of a paragraph and recognizes all inline tags.
-let rec parseChars acc input = seq {
+let rec parseChars acc input (ctx:ParsingContext) = seq {
 
   // Zero or one literals, depending whether there is some accumulated input
   let accLiterals = Lazy.Create(fun () ->
-    if List.isEmpty acc then []
-    else [Literal(String(List.rev acc |> Array.ofList))] )
+    if List.isEmpty acc then ([], ctx)
+    else ([Literal(String(List.rev acc |> Array.ofList), match ctx.CurrentRange with | Some(n) -> Some({ n with EndColumn = n.StartColumn + acc.Length }) | None -> None)], { ctx with CurrentRange = match ctx.CurrentRange with | Some(n) -> Some({ n with StartColumn = n.StartColumn + acc.Length }) | None -> None }) )
```
tpetricek: This looks like a super-long and hard to parse line - could you refactor it into something a bit more verbose, but a bit more readable?

src/FSharp.Markdown/MarkdownParser.fs
```diff
@@ -145,107 +146,128 @@ let (|Emphasised|_|) = function
...
+    yield LatexDisplayMath(String(Array.ofList body).Trim(), ctx.CurrentRange)
+    yield! parseChars [] rest ctx
 
   // Inline Latex inline math mode
   | DelimitedLatexInlineMath ['$'] (body, rest) ->
-    yield! accLiterals.Value
-    yield LatexInlineMath(String(Array.ofList body).Trim())
-    yield! parseChars [] rest
+    let (value, ctx) = accLiterals.Value
+    let ctx = { ctx with CurrentRange = match ctx.CurrentRange with | Some(n) -> Some({ n with StartColumn = n.StartColumn + 1 }) | None -> None }
```
tpetricek: Maybe these sort of operations could be helper functions, so that you can just say:

    let ctx = ctx.WithRange(incrementStart (0, 1) ctx.Range) // to increment start by `0` lines and `1` columns

Similar operations are done in a few places in the code, so a helper function might be useful (not necessarily exactly as I suggested above - but

src/FSharp.Markdown/MarkdownParser.fs
```diff
@@ -582,14 +605,16 @@ let (|EmacsTableBlock|_|) input =
...
   let regex =
     "^ {0,3}" // Up to three leading spaces
     + ">" // Blockquote character
     + "\s?" // Maybe one whitespace character
     + "(.*)" // Capture everything else
   let match' = Regex.Match(line, regex)
-  if match'.Success then Some (match'.Groups.Item(1)).Value
+  if match'.Success then 
+    let group = match'.Groups.Item(1)
+    Some (group.Value, { n with StartColumn = n.StartColumn + group.Index; EndColumn = n.StartColumn + group.Index + group.Length })
```
tpetricek: Hmm, looking at this, maybe my idea with helper functions is not that simple - the calculations on ranges seem to be pretty irregular. So feel free to say that this was a silly suggestion!

tpetricek:
@AviAvni Thanks very much for the revised version, it is a fantastic work! It's great to see how this is turning into a great F# Formatting improvement.

I think we are almost there - I made a few more small suggestions, but I think this is now almost at a stage where we can merge the PR.

I added a few nitpicking comments about the naming (because that's the bit which we cannot change once we merge the PR), but aside from that, there are really just smaller things.

Thanks a lot for putting all the effort into this - it is a fairly big change, so I'm sorry it is taking that long, but once it's in, it will be an amazing improvement :) :heart:

ljw1004:
+1

docs/content/codeformat.fsx
AviAvni: I didnt removed the parameter from the method declaration when I looked at all scrpts to see if I need to change something I saw error on this line so I change it

AviAvni:
@tpetricek @ljw1004 thanks for the review I made the changes you asked
you can see the comment about the issue of removing the parameter from CodeFormat.CreateAgent

The the issue of adding helper functions for manipulating MarkdownRange and ParsingContext
when I first implement it also to me it look like I need something like that but after I finish the implementation also like you think maybe it's ok like that 

any way I thought that after finish with this I'll try to implement the parser with parser combinator library then handling the ranges will lbe more simple so to me it doesn't matter

thanks for helping and letting me do the implementation it's a great library
and it's great that I can help both of you

tpetricek:
This is now released as version `3.0.0-beta01` (merged into #417, but not in `master` yet).

@AviAvni Thank you very much for working on this (and for patience with me :-))

@ljw1004 It would be great if you could try the beta version and let us know if everything works as expected. The new version also has named fields on discriminated union cases, which might make the VS interop nicer.

PR #417, opened 2016-08-01
Version 3

tpetricek:
This is now released as a beta, because it has quite large breaking changes. We'll make it master sooner or later.


src/FSharp.Formatting.Razor/Main.fs
```diff
@@ -0,0 +1,171 @@
...
+///    (if not specified, we use the currently loaded assemblies).
+///  - `sourceFolder` and `sourceRepo` - When specified, the documentation generator automatically
+///    generates links to GitHub pages for each of the entity.
+///  - `publicOnly` - When set to `false`, the tool will also generate documentation for non-public members
+///  - `libDirs` - Use this to specify additional paths where referenced DLL files can be found
+///  - `otherFlags` - Additional flags that are passed to the F# compiler (you can use this if you want to
+///    specify references explicitly etc.)
+///  - `urlRangeHighlight` - A function that can be used to override the default way of generating GitHub links
+///
+type RazorMetadataFormat private() =
```
matthid: I'm not exactly happy with the naming of this new module. We now have `MetadataFormat` `RazorMetadataFormat` and everything seems to be called `Generate`...
As we are already breaking we should try to find a better naming.

Ideas?
/cc @tpetricek @Krzysztof-Cieslak @cloudRoutine (because of https://github.com/tpetricek/FSharp.Formatting/issues/421) 

I think we are closing in on that Pipeline idea

For Documentation-Metadata we have (because of https://github.com/tpetricek/FSharp.Formatting/pull/425) 
 MetadataFormat: xml + dll -> DocumentationModel
 (Razor)MetadataFormat: DocumentationModel + Templates -> Html

For Literate it is rather similar I guess.

matthid:
I'm not sure if anyone is doing an older release. Therefore I'll merge this for now (I created a version2 branch just in case we need it).

@tpetricek We are now ready to move the project. [Links in API-Reference](http://tpetricek.github.io/FSharp.Formatting/reference/fsharp-codeformat-codeformat.html) are already broken. The rest will take some time anyway (there will always be the one outdated link).

PR #425, opened 2016-10-04
Remove razor

Krzysztof-Cieslak:
Trying to remove Razor dependency from core of F#.Formatting, and then add same functionality on top of it. It will enable using different rendering systems in the future.

Targeting `version3` branch, as it'll contain breaking changes.


matthid:
We should probably do this. Thanks for looking into this!

Change looks ok to me, I think we can do better on the generator type definition. Maybe with type abbreviations or records?

tpetricek:
It's great to see this happening. I think having "core library" with additional layers is a really good strategy.

I think `generator` is a good easy way to do this, though if I had infinite amount of time, I would prefer to do this in a way that does not involve callbacks. So as opposed to writing:

    ProcessDocument(file, template, fun html template replacements -> 
      Razor.ApplyTemplate(html, replacements, template))

I think it would be much nicer to write:

    let doc = ParseDocument(file)
    Razor.ApplyTemplate(doc.Html, doc.Replacements, template))

This is a more significant change - because all methods now need to _return_ result (and we need to define nice types for this result) rather than calling generator on their own.

But I think, in the long run, this really enables more uses of the library. Callbacks are framework smell :-).

That said, I'm happy to accept the PR as it is, because I think it's a good move.

Krzysztof-Cieslak:
As suggested by @matthid - using record for generator function input.

Krzysztof-Cieslak:
MetadataFormat returns F# record (as Tomas prefers ;) ). Will do the same for Formatting tomorrow :)

Krzysztof-Cieslak:
OK... I *think* it's ready

Fixed CommandLineTool and Tests by using Razor API build on top of my changes (it has exactly the same API as previously Core functionality). I guess in the future someone should move changed tests to Razor.Test project and add true unit tests for Core API.

Krzysztof-Cieslak:
Hmmm... looks like it's passing tests but failing on `GenerateDocs` task (which makes [breaking changes in  API] sens if you're dogfooding and use build version to gen its own docs )

cloudRoutine:
@tpetricek @matthid any chance one of you could take a look at this? I tried to figure out why the build was failing but I had no luck with it.

I know there have been issues with the VFPT core in the past, but I can publish tailored versions of it to https://ci.appveyor.com/nuget/fsharp-editing like I did to get the latest version using FCS 8.0.0 in https://github.com/Krzysztof-Cieslak/FSharp.Formatting/pull/1/commits 

It'd be great if we could get FSharp.Formatting ready to run on netcore, upgrade the framework targets to 4.6.1, adjust the src for api compatibility, etc.

I can take care of whatever you guys need on the VFPT side.

matthid:
Problem was that we now have a single instead of two caches, and `Literate` and `MetadataFormat` generation are using different namespaces (which might influence the razor compilation and was therefore forbidden) ... 

While strictly speaking it would be better to leave the exception and add the optional `namespaces` argument to all functions, I think its better for now to make it work and print a warning that performance could be improved...

Anyway it was good to notice (and fix) this breaking change.

Krzysztof-Cieslak:
Thanks a lot for merging, sorry I haven't done last fixes / suggestion but I'm pretty busy with some other stuff recently.

matthid:
@Krzysztof-Cieslak No problem, let me know if you need a release (for example to play with Fornax) ;)
Will do those minor things later 👍

cloudRoutine:
@matthid you're my hero 🎊 
this has unblocked a bunch of stuff I've wanted to do with `FSharp.Formatting`

matthid:
really @Krzysztof-Cieslak is the hero here ...

cloudRoutine:
he already knows how much I appreciate this 😉

PR #433, opened 2017-04-10
Move to fsprojects

tpetricek:
I chatted about F# Formatting with @matthid briefly at F# eXchange - given all the awesome recent work on Version 3 in #417, I think it wold make sense to move F# Formatting to fsprojects, along with the release of Version 3.

I think the right way to do this is to update the `gh-pages` in the old repository one last time before moving the project - that way, the current `gh-pages` content will point to the new URLs and I _think_ GitHub will arrange redirects from the old location.

This PR replaces the URL and we should merge it when we are ready to move the project to "fsprojects". Once it is merged & `gh-pages` are updated, let me know and I can do the transfer!


matthid:
Done (see https://github.com/tpetricek/FSharp.Formatting/commit/0c926c7edb32f25d59d35070b0944cf8a424c06f).
I think we are ready and we have no reason to wait for the version3 release either.

@tpetricek :shipit:

dsyme:
@tpetricek @matthid How close are we to a new release of FSharp.Formatting, update to use more recent versions of the compiler service?

matthid:
@dsyme I can only speak for myself. And for me it's a time problem:

 - My focus is on FAKE at the moment (and I started some huge changes there) which means working on Paket and FAKE -> FSF is falling behind.
 - It's still not clear (at least to me) how we handle references in the new dnc package based world
   On one hand it's a huge advantage that we can finally figure out our references everywhere (and don't need a global installation of some reference assemblies and some hacks to use them outside of msbuild)
    On the other hand it feels wrong (I discussed this with several people) to add "Paket.Core" to all projects which need to use FSC one way or another to get the required reference list.

The second point is another reason why I'm waiting a bit to see what others do.

To answer your question: I think there is already a build with a more recent FSC build available (no netcore support), you can use the AppVeyor feed (https://ci.appveyor.com/nuget/fsharp-formatting) to unlock yourself if needed.

Yes we could release that I guess (I wanted v3 to be the version with netcore support that's why I haven't done that yet).

Hope that clarifies the situation.

dsyme:
@matthid Great, thanks for the update.  I was just after an update really.  I'd like to make sure we move the ecosystem forward to more recent FCS versions.  I'm not blocked

Perhaps we should write a roadmap issue for FSharp.Formatting, or flush out the issue list (old suggestions, discussions etc.) so the priority work items can be seen

PR #446, opened 2017-08-08
Specify VisualStudioVersion so it finds Microsoft.FSharp.targets

conniey:
This fixes the "The value "" of the "Project" attribute in element <Import> is invalid. Parameter "path" cannot have zero length." error because it cannot locate Microsoft.FSharp.targets.
The root cause appears to be that the [VisualStudioVersion is not being set at the appropriate level](https://stackoverflow.com/questions/38343130/error-msb4102-the-value-of-the-project-attribute-in-element-import-is-in)

I copied the contents from: [FSharp.CodeFormat.fsproj#L37-L66](https://github.com/fsprojects/FSharp.Formatting/blob/master/src/FSharp.CodeFormat/FSharp.CodeFormat.fsproj#L37-L66) to set the VS version to 2015.

/cc: @matthid


matthid:
Yes that works, in other repros we used the fsharp.compiler.tools package.
I thought we were already using it here as well (and was a bit puzzled on your error report).
This should work as well, thanks!

conniey:
@matthid Based on your feedback, I modified each project to use FSharp.Compiler.Tools instead. Does this work better?

matthid:
The build is failing because one project uses nuget for some reason... It should work when migrating it to paket (ie packages.config -> paket.references -> paket install, maybe some manual cleanup of the fsproj)

Thanks! Let me know if you run into trouble.

conniey:
I fixed it now.. and both builds appears to be passing. :)

conniey:
@matthid Hey, does this work for you?

matthid:
Yes absolutely, sorry :)

Issue #453, opened 2017-12-24
publish the 3.0.0-beta08 nupkg to Nuget.org


jackfoxy:
Only the beta01 is available for download. 

Among other problems the nuspec points to an old version of FCS (2.0.0.6) and it contains an older version of FSharp.Formatting.fsx.

matthid:
We publish prereleases to the appveyor feed (`https://ci.appveyor.com/nuget/fsharp-formatting`). Just add that for now.
Publishing to appveyor is easier for me.

jackfoxy:
3.0.0-beta08 is already very good and fixes https://github.com/fsprojects/FSharp.Formatting/issues/364 (for me). Just by incorporating the latest FCS it probably fixes some other issues, too. 

My github.io pages finally have tooltips over the F# samples again.

This is ready for wider beta testing. It would be useful to publish some guidance on the changes people must make to their generate.fsx files, so they don't have to dig up the generate.fsx file in this repo.

Putting it on Nuget will encourage more beta testers.

wallymathieu:
I've done a pull request to fix issue with rendering in F#+ https://github.com/gusty/FSharpPlus/pull/79 , it looks kind of weird to reference an outside feed.

Having a migration doc would also be nice.

jackfoxy:
+1 to migration doc. I had to piece it together myself. I won't have time myself to put together such a doc in the near future. Maybe in a couple of weeks.

wallymathieu:
But I'm not sure that helps

jackfoxy:
For myself, and probably for most people, it is the changes necessary for to the generate.fsx file that many people have from Project Scaffold.

wallymathieu:
Where did you integrate the beta?

wallymathieu:
So that we can perhaps compile a list of changes needed?

jackfoxy:
Both of these projects source FSharp.Formatting 3.0.0-beta08 with Paket from https://ci.appveyor.com/nuget/fsharp-formatting

https://github.com/jackfoxy/DependentTypes

https://github.com/jackfoxy/FsRegex

in both git histories you will find F#.Formatting beta 8 commits. The primary change is to docsrc/tools/generate.fsx

PR #459, opened 2018-02-04
Adding markdown with a bit of info about how to upgrade

wallymathieu:
Open question:

- [x] Where should this information be linked from?

I've used feedback from @jackfoxy on what should be part of such a upgrade path https://github.com/fsprojects/FSharp.Formatting/issues/453

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
  - Fix some links on the website - https://github.com/fsprojects/FSharp.Formatting/pull/458
  - Another link on the website - https://github.com/fsprojects/FSharp.Formatting/pull/454
  - Support highlighting for paket.dependencies `storage` keyword - https://github.com/fsprojects/FSharp.Formatting/pull/451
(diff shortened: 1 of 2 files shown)


matthid:
I guess one location would be the release notes. We could add that link for every beta until we do the release.

jackfoxy:
The current upgrade link in release notes 404 http://fsprojects.github.io/FSharp.Formatting/upgrade_from_v2_to_v3.html

wallymathieu:
Thanks @jackfoxy !

wallymathieu:
@matthid do we need to generate the documentation from the v2 branch (in order to avoid docs from v3 to be shown)?

matthid:
Good question... We have no general solution to that problem ...

wallymathieu:
I can cherry pick d35fb84651e0eee3469f879d7792bcda43b5b4bf to the v2 branch in order to regenerate docs

matthid:
Honestly, I'm OK either way (ie generating docs from v3 or cherry picking into v2).

wallymathieu:
I've tried regenerating the docs, and it looks like it gets further when I start on #460, I'll try to diagnose what goes wrong

PR #464, opened 2018-03-12
Only replace ampersands if they don't belong to another HTML entity

stroborobo:
Inline ampersands were previously replaced by `&amp;`, even if they already belonged to another entity.

For example this:

```
Type: **'T -&gt; 'U**
```

Became this:

```
<p>Type: <strong>'T -&amp;gt; 'U</strong></p>
```

[But it should have been this](http://spec.commonmark.org/dingus/?text=Type%3A%20**%27T%20-%26gt%3B%20%27U**%0A):

```
<p>Type: <strong>'T -&gt; 'U</strong></p>
```

I'm fairly new to F#, so I hope this is PR is ok.


wallymathieu:
What do you think @matthid ?

matthid:
I'm not sure, what does the markdown spec say to this?

stroborobo:
@matthid 

> All valid HTML entity references and numeric character references, except those occuring in code blocks and code spans, are recognized as such and treated as equivalent to the corresponding Unicode characters. Conforming CommonMark parsers need not store information about whether a particular character was represented in the source using a Unicode character or an entity reference.

http://spec.commonmark.org/0.28/#entity-and-numeric-character-references

stroborobo:
Nice, thanks for merging, guys!

wallymathieu:
Thanks for the contribution @stroborobo ! 👍

PR #467, opened 2018-05-05
WIP: Adds support for attributes on modules, types and members

kblohm:
Hi,
i have added support for attributes declared on types, members and modules.
There are also some functions to support the System.ObsoleteAttribute so it is easier to mark stuff as obsolete in documentation.

There is some stuff i am not quite sure about:

* As of now i do not add attributes to function parameters etc. That would probably not be too hard in general, but MemberOrValue.TypeArguments is just a string list right now, so its not really compatible.
* I have also not modified any templates yet. There are functions to format and an attribute/ multiple attributes, but i did not want to break all the existing templates. I am not sure whats the right approach here.
* If there are any plans to add attributes into the types list of a namespace/module, there might be some naming conflicts later on. But i am not sure if that is too important.
* I added the attributes for members right on the Member type and not on MemberOrValue. I am not sure what u prefer here.

Some side notes:
* the editorconfig enforces 4 spaces but alot of the files use 2. That makes it kinda hard to work with ;)
* is it just me or does the intellisense not work in vscode? I had to use vs. Not sure what i am doing wrong here :(

### DIFF
--- a/misc/templates/reference/module.cshtml
+++ b/misc/templates/reference/module.cshtml
@@ -33,10 +33,28 @@
 
 <h1>@Model.Module.Name</h1>
 <p>
+    @if (Model.Module.IsObsolete)
+    {
+        <div class="alert alert-warning">
+        <strong>WARNING: </strong> This API is obsolete
+        <p>@Html.Encode(@Model.Module.ObsoleteMessage)</p>
+        </div> 
+    }
   <span>Namespace: @Model.Namespace.Name</span><br />
   @if(Model.ParentModule.Exists())
   {
   <span>Parent Module: <a href="@(Model.ParentModule.Value.UrlName).html">@Model.ParentModule.Value.Name</a></span>
+  }
+  @if (Model.Module.Attributes.Any())
+  {
+      <span>
+          Attributes:<br />
+          @foreach (var attr in Model.Module.Attributes)
+          {
+                @Html.Encode(@attr.Format())<br />
+          }
+          
+      </span>
   }
 </p>
 <div class="xmldoc">
--- a/src/FSharp.MetadataFormat/Main.fs
+++ b/src/FSharp.MetadataFormat/Main.fs
@@ -14,7 +14,7 @@ open System.Text
 open System.IO
 open System.Xml
 open System.Xml.Linq
-
+/// Represents a comment attached to F# source code
 type Comment =
   { Blurb : string
     FullText : string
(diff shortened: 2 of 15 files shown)


kblohm:
> Not sure what you mean there? How will changing the templates break stuff? Or do you mean using the templates with older versions of the library? I guess we don't really need to be compatible in that way, because people usually copy (and modify) the templates or they use the ones bundled with the nuget package (but again it will contain a compatible version).

It would not really break stuff but someone who uses the default template might be surprised. So the general question i should have asked is probably: Should a signature with attributes (not just obsolete) be the default in a template? Should it be optional? If none of the two we can leave it as is. Otherwise i can try to think of a reasonable default in a template.

> Again I'm not sure I understand.

After reading that again, i did not really understand it myself anymore ;P. 
So let me rephase that:
If i define an attribute in my API, is it currently listed under any of the types so i can generate documentation for it? If not, is this planned? Because if it is planned, it might be confusing what attribute property is for the attributes my types consume (for example the obsolete one) and what are the attributes i declared myself and want to have documented. Does that make sense?

> Feed free to change to 2 I guess ;)

I can also change the file to use 4 if u dont mind. I like 4;)

matthid:
> It would not really break stuff but someone who uses the default template might be surprised.

If people do not copy but reference our template, they should expect it to change and they decided to opt-in into new features (at least that's my standpoint on this)

> If i define an attribute in my API, is it currently listed under any of the types so i can generate documentation for it?

Yes it is a regular type?

> Because if it is planned, it might be confusing what attribute property is for the attributes my types consume (for example the obsolete one) and what are the attributes i declared myself and want to have documented.

In the razor-template you can filter out and write custom code to handle that if it is useful for someone.

> I can also change the file to use 4 if u dont mind. I like 4;)

In general yes, but please in a separate PR for this. remembering how some modules are written I expect some modules to go "quite a bit to the right" and might need some additional refactoring if you really want to start this "mess" ;)

Krzysztof-Cieslak:
Seems to be working for me, but initialization/loading of projects at startup is bit slow.

screenshot from 2018-05-05 20-13-37

matthid:
> As far as i can see u just cant tell if a type is a module or a class or an attribute right now. 

I think basically the logic is: If some base class in the hierarchy is `System.Attribute`...

> do u know how i can get proper intellisense for the razor templates?

Don't waste too much time on this. It worked with a hack on older Visual Studio versions but honestly I'd like to throw Razor away eventually. My current thinking is to replace (= mark obsolete) Razor short/mid-term with https://mangelmaxime.github.io/Fulma/#fulma
In my current thinking templates would be specially written `.fsx` files compiled and executed by fsharp.formatting

Krzysztof-Cieslak:
> Don't waste too much time on this. It worked with a hack on older Visual Studio versions but honestly I'd like to throw Razor away eventually.

Maybe move Razor to separate NuGet package so people that want it can continue using it? It should be fairly simple since all Razor code is in its own project? While I love the idea of removing it I think it may be bit too big change for many people.

matthid:
@Krzysztof-Cieslak 

> My current thinking is to replace (= mark obsolete) Razor 

Yes it will still be there but it won't be the first choice anymore

matthid:
@Krzysztof-Cieslak Even more important would be to get rid of the requirement that we have to provide all dependencies to FCS. This would remove quite some churn from the ecosystem

matthid:
> Maybe there is some interest in cooperating with them 

Maybe, but I'd assume creating the yaml is actually the hardest part, writing the template is not too hard and FSF provides full flexibility. We would probably loose tool tips when hovering over F# code.

> I have added IsObsolete and ObsoleteMessage on every type with attributes.

Answering in-line

src/FSharp.MetadataFormat/Main.fs
```diff
@@ -24,6 +24,91 @@ type Comment =
...
+          FullName = fullName
+          ConstructorArguments = constructorArguments
+          NamedConstructorArguments = namedConstructorArguments }
+  static member private RemoveAttributeSuffix(name:string) =
+      let attributeSuffix = name.LastIndexOf("Attribute");
+      if attributeSuffix = -1 then name else name.Remove(attributeSuffix,9)
+
+  /// Gets a value indicating whether this attribute the System.ObsoleteAttribute
+  member x.IsObsoleteAttribute =
+    x.FullName = "System.ObsoleteAttribute"
```
matthid: I'm fine with this

src/FSharp.MetadataFormat/Main.fs
```diff
@@ -24,6 +24,91 @@ type Comment =
...
+
+  /// Returns the obsolete message, when this attribute is the System.ObsoleteAttribute. When its not or no message was specified, an empty string is returned
+  member x.ObsoleteMessage =
```
matthid: I'd personally move this up

src/FSharp.MetadataFormat/Main.fs
```diff
@@ -24,6 +24,91 @@ type Comment =
...
+
+  /// Formats the attributes using the Name. Removes the "Attribute"-suffix by default e.g [<Obsolete>] instead of [<ObsoleteAttribute>]
+  static member FormatAttributes(attributes:seq<Attribute>, ?removeAttributeSuffix:bool) =
+    String.Join("\n", attributes |> Seq.map (fun a -> a.Format(?removeAttributeSuffix = removeAttributeSuffix)))
+
+  /// Formats the attributes using the Name. Removes the "Attribute"-suffix by default e.g [<Obsolete>] instead of [<ObsoleteAttribute>]
+  static member FormatAttributesFullName(attributes:seq<Attribute>, ?removeAttributeSuffix:bool) =
+    String.Join("\n", attributes |> Seq.map (fun a -> a.FormatFullName(?removeAttributeSuffix = removeAttributeSuffix)))
+  /// Tries to find the System.ObsoleteAttribute and return its obsolete message
+  static member TryGetObsoleteMessage(attributes:seq<Attribute>)=
```
matthid: Again, I'd move that up

src/FSharp.MetadataFormat/Main.fs
```diff
@@ -58,21 +143,41 @@ type MemberKind =
...
+    Attribute.FormatAttributes(x.Attributes, ?removeAttributeSuffix = removeAttributeSuffix)
+  /// Formats the attributes using the Name. Removes the "Attribute"-suffix by default e.g [<Obsolete>] instead of [<ObsoleteAttribute>]
+
+  member x.FormatAttributesFullName(?removeAttributeSuffix:bool) =
+    Attribute.FormatAttributesFullName(x.Attributes, ?removeAttributeSuffix = removeAttributeSuffix)
+     
+  /// Gets a value indicating whether this member is obsolete
+  member x.IsObsolete =
+    x.Attributes
+    |> Seq.exists (fun a -> a.IsObsoleteAttribute)
```
matthid: Fine by me

src/FSharp.MetadataFormat/Main.fs
```diff
@@ -58,21 +143,41 @@ type MemberKind =
...
+
+  /// Returns the obsolete message, when this member is obsolete. When its not or no message was specified, an empty string is returned
+  member x.ObsoleteMessage =
+    Attribute.TryGetObsoleteMessage(x.Attributes)
```
matthid: In-line the required code here.
I see that this is not really "clean" but I have no suggestion on how to improve.

Strictly speaking I'm fine with either way. Note that extensions only work when using regular C# extension methods and might lead to problems with the razor template, so I guess leaving this a member is as good as it gets.

Krzysztof-Cieslak:
> Even more important would be to get rid of the requirement that we have to provide all dependencies to FCS. This would remove quite some churn from the ecosystem

One can only dream ;-)

matthid:
> One more question: am i supposed to HTML escape comments myself?

The problem is that this are XML-comments if you do not write `<summary></summary>` it is implicit in F#, but it still is xml. Therefore I think it doesn't like `<`

misc/templates/reference/module.cshtml
```diff
@@ -33,11 +33,28 @@
 
 <h1>@Model.Module.Name</h1>
 <p>
+    @if (Model.Module.IsObsolete)
+    {
+        <div class="alert alert-warning">
+        <strong>WARNING: </strong> This API is obsolete
+        </div>
```
matthid: Should we include the obsolete message (as you have already implemented that)?

kblohm: We can. The actual message is also listed under the attributes. So if u hover over a method or click into a type/module u can see it there. But i can also include it here.
For Fake i would definitely include it as there are so many obsolete types. But in most APIS there are not that many.

matthid:
So this is ready?

kblohm:
Sorry, that should be it.
If i escape it, the HTML is fine but vs-tooltips are not.. that is kind of dumb :p

matthid:
> Sorry, that should be it.

No need to be sorry I just couldn't wait to merge ;)

> If i escape it, the HTML is fine but vs-tooltips are not.. that is kind of dumb :p

Yeah maybe an fsharp.formatting bug ;)

Thanks a lot for this - this is indeed very useful!

Issue #402, opened 2016-04-27
lineNumbers can't be turned off with the command line tool


AnthonyLloyd:
Target "Posts" (fun _ ->
    FSharpFormatting.run "literate --processDirectory --inputDirectory  posts --outputDirectory _posts --lineNumbers false" 
)

Sorry I've not been able to investigate further but thought I would report

tpetricek:
A PR that makes this possible would be great.

AnthonyLloyd:
Looks like a problem is with CommandLineParser gsscoder/commandline#115.
They said they won't fix until 2.0 but I'm not sure it is really being maintained.

dsyme:
This was completed, thanks

PR #481, opened 2018-10-14
CommandTool: replace lineNumbers with noLineNumbers.

### LINKED ISSUE #479: Can’t avoid line numbers with command line tool
The command line tool has `—lineNumbers` which defaults to `true`. Unfortunately that means you effectively can not turn it off looks like  - because CommandLineParser interprets absence of the arg as the default, and presence of the arg as true also. 

Lmk if I missed anything here. 

If not, any preference for changing the default to false, OR changing the switch to `—noLineNumbers`? Happy to PR either.

kurtschelfthout:
Fixes #479 and #402.


wallymathieu:
We could close and reopen the pull request in order to get appveyor to rebuild.

teo-tsirpanis:
> Not sure what's going on with the appveyor build, but it looks unrelated to this PR. Some kind of paket restore problem.

@kurtschelfthout, [that's funny](https://help.appveyor.com/discussions/problems/16818-appveyor-often-fails-without-explanation).

PR #483, opened 2018-10-27
Fix Literate nuget build

thinkbeforecoding:
@matthid there was a little mistake in the build of the nuget for FSharp.Literate. It was actually using the nuspec of FSharp.Formatting 😬 

This PR is fixing this hopefully.


matthid:
Thanks :)

PR #485, opened 2019-01-23
FEAT: add optional 'markDownComments' & 'publicOnly' command arguments

stijnmoreels:
Proposal, is this enough to just add both optional arguments to the command line tool?


wallymathieu:
Does it work locally when you try it out?

stijnmoreels:
> Does it work locally when you try it out?

Well, I tried it with an existing project; but I don't see the same result for the `dllFiles`. But I guess that's just a matter of different versions. 
Not sure otherwise to test this? I see both arguments in the "help" screen though.

stijnmoreels:
Ok, now it seems that the arguments are passed in.

But I don't think that I get the expected results? Is there actually any support for:

```xml
<summary>
</summary>
<param name="x">This is a parameter</param>
```

parameters?

wallymathieu:
I'm unsure about that. Perhaps @matthid knows better?

PR #489, opened 2019-04-06
Fix building NuGet packages on unix-based systems

iNecas:
When trying to build nuget package on Linux, I've hit the following
issue:

```
$ fake run build.fsx -t NuGet -s
Script reported an error:
-> BuildFailedException: Target 'NuGet' failed.
-> One or more errors occurred. (Could not find a part of the path
'NuGet/FSharp.Formatting.3.0.0-beta13.nuspec'.)
-> Could not find a part of the path
'NuGet/FSharp.Formatting.3.0.0-beta13.nuspec'.
```

The problem seems to be, while Fake's NuGet generates files in `NuGet`,
the repository currently uses `nuget`, which on Linux are two completely
different dirs.

Unifying on `NuGet` seems to fix the issue for me.

My editor also auto-fixed some whitespace issues: I hope nobody minds much. I like the version without trailing spaces a bit better :)


matthid:
Thanks for the fix. Isn't this a bug in FAKE as well?

iNecas:
But in what sense?

PR #492, opened 2019-04-12
update FCS and other dependencies, use .NET Core FAKE

dsyme:
This 
* updates to FCS latest
* adjusts `build.fsx` to use .NET Core FAKE
* adjusts `build.fsx` to use inline paket dependencies, which I think are clearer and more isolated (the use of FAKE doesn't infect the `paket.dependencies` file)


tests/FSharp.CodeFormat.Tests/FSharp.CodeFormat.Tests.fsproj
```diff
@@ -6,7 +6,7 @@
     <OutputType>Library</OutputType>
     <OutputPath>$(SolutionDir)\tests\bin</OutputPath>
     <DocumentationFile>$(OutputPath)\$(AssemblyFile).xml</DocumentationFile>
-    <AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
```
dsyme: Ugh not sure what made this change, was not intended

dsyme:
Looks like Yaaf Scripting needs to be updated

```
 C:\GitHub\dsyme\FSharp.Formatting\paket-files\matthid\Yaaf.FSharp.Scripting\src\source\Yaaf.FSharp.Scripting\YaafFSharpScripting.fs(66,35): error FS0039: The field, constructor or member 'Create' is not defined.
Fsc: C:\GitHub\dsyme\FSharp.Formatting\paket-files\matthid\Yaaf.FSharp.Scripting\src\source\Yaaf.FSharp.Scripting\YaafFSharpScripting.fs(281,25): error FS0072: Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.
```

dsyme:
@matthid Update is here: https://github.com/matthid/Yaaf.FSharp.Scripting/pull/13

dsyme:
I updated this PR - please take a look

1. `generate.fsx` is now a F# Interactive script, not a FAKE script.  This reduces dependency collisions as we update `FSharp.Compiler.Service`

2. documentation generation bootstraps by using the DLLs from `/tests/bin/net461` rather than via a copy process

3. documentation generation now only runs on Windows.  I think this is OK - it currently depends too much on .NET Framework to bother with anything else until we make documentation generation seamless on .NET Core tooling (eg. without generate.fsx scripts).

wallymathieu:
I think it looks ok. Using `+"/"+` over `@@` makes it slightly clearer. My primary usage of the lib is on appveyor.

matthid:
> documentation generation now only runs on Windows. I think this is OK - it currently depends too much on .NET Framework to bother with anything else until we make documentation generation seamless on .NET Core tooling (eg. without `generate.fsx` scripts).

I for myself always tried to keep it working on mono. It's a bit sad if we can publish/release all F# open source projects only on windows but call ourself cross-platform. Can you clarify what exactly doesn't work anymore?

dsyme:
> I for myself always tried to keep it working on mono. It's a bit sad if we can publish/release all F# open source projects only on windows but call ourself cross-platform. Can you clarify what exactly doesn't work anymore?

I'll fix it up.  But the point is really that we need to focus on a viable cross-platform documentation generation story, and the use of `generate.fsx` scripts to drive documentation generation is very broken going forward.  We should replace the use of `generate.fsx` with the use of the command line tool.

Indeed do you know if `DogFoodCommandTool` is actually a full replacement for the documentation generation?  If so we can just get rid of `generate.fsx` straight away and replace with `DogFoodCommandTool`.

matthid:
You can get very very far with the command tool, but obviously, if you are doing heavy customization or using FSharp.Formatting classes/types in your script (or overwrite them) you will have a hard time switching to the tool.
I'd say the command-line tool will cover `>90%` use cases (which is not covered by any hard facts).

For example with code, you can achieve a very customized traversal of the "markdown AST" and handle some specific edge cases yourself or integrate external libraries like "Prism" instead of using `FSharp.Formatting` defaults for code-view.

dsyme:
> > Indeed do you know if DogFoodCommandTool is actually a full replacement for the documentation generation? 

> You can get very very far with the command tool.

Right, though I'm meaning specifically for doc generation in this repo and normal-case doc generation in ProjectScaffold repos.

> if you are doing heavy customization or using FSharp.Formatting classes/types in your script ...

Yes, I'm only really talking about what we really recommend in ProjectScaffold (which already uses the command line tool).  What people do with the API is up to them (though as we've discussed I wouldn't really recommend scripting with it - the heavy dependency chains of FSharp.Formatting libraries is always going to make in-process execution of these libraries to do reflection over other .NET Libraries a bit dodgy )

dsyme:
OK I think this is good to go in.

dsyme:
> I for myself always tried to keep it working on mono. 

This is fixed, I'll make a separate PR to just remove `generate.fsx`

iNecas:
Hah, I was just fighting with getting `FSharp.Formatting` working with latest `FSharp.Compiler.Service` and for a newcomer, it took me a while to find out why the things is not working even when the code was compiling, to find the dependency collisions in generate.fsx and FAKE.

+1 for better cross-platform story based on dotnet core.

Also thanks for fixing the generate.fsx on mono for now: I believe not having that working would hurt potential F# adoption in general: I guess quite of lot of people exploring the F# ecosystem will come across this project early in the process (probably via ProjectScaffold) and it's not a good sign to learn it being broken outside of Windows platform.

paket.lock
```diff
@@ -1,13 +1,8 @@
...
   remote: https://api.nuget.org/v3/index.json
-    Argu (5.1)
-      FSharp.Core (>= 4.0.0.1) - restriction: >= net45
-      FSharp.Core (>= 4.3.2) - restriction: && (< net45) (>= netstandard2.0)
-      System.Configuration.ConfigurationManager (>= 4.4) - restriction: && (< net45) (>= netstandard2.0)
     CommandLineParser (1.9.71)
-    FAKE (5.8.4)
-    FSharp.Compiler.Service (25.0.1) - content: none, redirects: force
-      FSharp.Core (>= 4.1.18) - restriction: || (>= net45) (>= netstandard2.0)
+    FSharp.Compiler.Service (27.0.1)
```
iNecas: @dsyme  I see you had even `28.0` here before: what was to reason to downgrade?

dsyme:
> Also thanks for fixing the generate.fsx on mono for now: I believe not having that working would hurt potential F# adoption in general: I guess quite of lot of people exploring the F# ecosystem will come across this project early in the process (probably via ProjectScaffold) and it's not a good sign to learn it being broken outside of Windows platform.

Yes this is a structurally important library - a lot of my recent investigations on moving more of the tooling chains on to .NET Core lead back here.  

(BTW I don't actually know much about Razor so I'd really like help on getting rid of that .NET Framework dependency...)

thinkbeforecoding:
It seems using razor for netstandard2.0 is simpler than older versions…see the 2nd answer to his SO question (didn’t try yet…)

https://stackoverflow.com/questions/38247080/using-razor-outside-of-mvc-in-net-core

jeremie chassaing / thinkbeforecoding

De : Don Syme
Envoyé le :samedi 13 avril 2019 00:44
À : fsprojects/FSharp.Formatting
Cc : Subscribed
Objet :Re: [fsprojects/FSharp.Formatting] update FCS and other dependencies,use .NET Core FAKE (#492)

Also thanks for fixing the generate.fsx on mono for now: I believe not having that working would hurt potential F# adoption in general: I guess quite of lot of people exploring the F# ecosystem will come across this project early in the process (probably via ProjectScaffold) and it's not a good sign to learn it being broken outside of Windows platform.
Yes this is a structurally important library - a lot of my recent investigations on moving more of the tooling chains on to .NET Core lead back here.
(BTW I don't actually know much about Razor so I'd really like help on getting rid of that .NET Framework dependency...)

Issue #475, opened 2018-08-01
Consider publishing the beta releases to NuGet.


teo-tsirpanis:
Since that version 3.0.0-beta01 is on NuGet, it gives the impression that there are no other releases after that, despite the fact that they exist on Appeyor, where they are not that easy to find.

matthid:
Yes the workflow is easier on appveyor... but we might need to document it better

tpetricek:
Just to follow-up on this - I was at open F# and two people there asked about our plan to release new non-beta release of F# Formatting on NuGet...

I have not been very closely following the developments, but what do we need before we can do that? One of the two said they're very happily using the beta and have no issues with it, so it sounds like it would be nice to have a new release sometime soon...

matthid:
> but what do we need before we can do that

What exactly? 

* Releasing existing beta releases to nuget is just a matter of "doing it" the good thing about appveyor is that I can "just" commit and don't need to run the release process locally (and is therefore faster).
* Releasing a "stable" version is another subject, there are a lot of ideas floating around but nothing has been finished yet. Minimal requirement (imho) for a stable v3 release is to support netcore/netstandard-dlls as well as running on netcore/netstandard. Ideally we use the breaking change to no longer bundle "razor" and instead use something else by default.

wallymathieu:
I could help with publishing beta to nuget if needed.

tpetricek:
Thanks for the quick reply @matthid and for the offer @wallymathieu!

I got the impression (from what I heard at open F#) that netcore/netstandard was already working quite well?

As for removing Razor dependency, I completely support that - though perhaps we could get away with not replacing it with anything initially and letting people add whatever they want as a separate component? (Replacing it completely sounds like quite a lot of work...).

thinkbeforecoding:
Releasing only fsharp.literate as a nuget for netstandard2 would be awesome.
Then using fable react templating…

jeremie chassaing / thinkbeforecoding

De : Tomas Petricek
Envoyé le :mardi 2 octobre 2018 15:38
À : fsprojects/FSharp.Formatting
Cc : Subscribed
Objet :Re: [fsprojects/FSharp.Formatting] Consider publishing the betareleases to NuGet. (#475)

Thanks for the quick reply @matthid and for the offer @wallymathieu!
I got the impression (from what I heard at open F#) that netcore/netstandard was already working quite well?
[...]

PR #493, opened 2019-04-12
remove beta tag and prepare to push nugets

dsyme:
This 

1. removes the beta tag (the packages are already widely used even in ProjectScaffold)
2. adds push-to-nuget.org to the Release steps

Regarding beta: the main thing missing is "run on .NET Core only".  As soon as that is added we can re-release and bump to 4.x
 
See also https://github.com/fsprojects/FSharp.Formatting/issues/475#issuecomment-482644142


dsyme:
@matthid or @forki Could you please give the nuget organization "fsprojects" access to nuget package "FSHarp.Formatting" and "FSharp.Literate", thank you

matthid:
Honestly, I wouldn't really remove the beta tag.
This leads us into the unfortunate situation to release breaking changes and v4 v5 "rapidly" without really adding new features.
Releasing v3 would indicate to users that the API is stable, which it is not.

dsyme:
> Honestly, I wouldn't really remove the beta tag.
> 
> This leads us into the unfortunate situation to release breaking changes and v4 v5 "rapidly" without really adding new features.
> 
> Releasing v3 would indicate to users that the API is stable, which it is not.

I really don't think that's a problem - like FCS we need to refine this API, and I really don't mind if we keep breaking it until version 400.   We just need to SemVer our way through this.

matthid:
While it is compatible with semver to always increase major on every release this is not how things should be. And it is not how you embrace people to use your library. I'm not really happy with how FCS is released to be honest.

Some quotes from semver.org:

> ### Doesn’t this discourage rapid development and fast iteration?
> Major version zero is all about rapid development. If you’re changing the API every day you should either still be in version 0.y.z or on a separate development branch working on the next major version.

> ### If even the tiniest backwards incompatible changes to the public API require a major version bump, won’t I end up at version 42.0.0 very rapidly?

> This is a question of responsible development and foresight. Incompatible changes should not be introduced lightly to software that has a lot of dependent code. The cost that must be incurred to upgrade can be significant. Having to bump major versions to release incompatible changes means you’ll think through the impact of your changes, and evaluate the cost/benefit ratio involved.

dsyme:
@matthid I do understand.  But we need to iterate on this one.  

I've held back on upgrading FSharp.Formatting in _any_ repository because it's been in beta for several years.   So we have many repositories with an effective dependency on FCS 2.0.0.0 and so on.   FSHarp.Formatting is not a particularly important part of those repos - it's just used for doc generation, but it's causing a lot of stasis. I now understand it is adoptable, but yes, we need to iterate a bit on it.

I agree about FCS - we'd all love to make it binary-compatible, but we've much more iterating before we can do that.

dsyme:
BTW there are two things driving my work here:

- FsLab has dependencies on FSharp.Formatting and we need to start updating those.  Oddly, I'm more comfortable building FsLab on a SemVer FSharp.Formatting that uses iterative improvement and major version updates than on a pre-release beta.   

- The rapidly growing need to shift the "fsprojects" ecosystem to use .NET Core tooling, rather than Mono.  Again, I'm more comfortable sitting on SemVer FSharp.Formatting than some prerelease beta 

So I'm strongly willing to help push things along, but only if we can use iterative improvement here (not based around beta/prerelease tags, which tend to be upwards infective )

matthid:
Yeah, the release is now out anyway, just wanted to share my opinion on this. I guess I'm more conservative in that regard and had a fixed set of features in mind for v3.

dsyme:
> Yeah, the release is now out anyway, just wanted to share my opinion on this. I guess I'm more conservative in that regard and had a fixed set of features in mind for v3.

I know, sorry.  Can you put those in a ROADMAP.md?

dsyme:
Oh can you give "fsprojects" access to FSharp.Formatting.CommandLineTool please?  That package has not been pushed yet

matthid:
@dsyme Request is pending

dsyme:
OK, packages pushed

I will now adopt these latest packages for FsLab (see [`fsx2html` aka `FSharp.Literate.Scripting` package](https://github.com/fslaborg/FsLab/tree/master/src/fsx2html) which is used to process F# scripts to docs), and somehow push hard to get us entirely on .NET Core.  

I will would also like to get a grip on the other various places where F# literate scripting and F# file watching is being implemented, including taking a good look at FCSWatch and also the various tools mentioned in https://github.com/Microsoft/visualfsharp/issues/6484

Issue #511, opened 2020-04-14
generate script execution in FSI is broken after 35.0.0 Update from FSharp.Compiler.Service


kMutagene:
Hi there, it is great that we can move to full .net core doc generation with generate scripts again. I just noticed that 4 days ago, the [FSharp.Compiler.Service](https://www.nuget.org/packages/FSharp.Compiler.Service/) got an update that broke doc generation again. The error is the following:

`System.TypeLoadException: Could not load type 'ParsedInput' from assembly 'FSharp.Compiler.Service, Version=35.0.0.0, Culture=neutral, PublicKeyToken=null'.`

When pinning a dependency to the previous version:

`nuget FSharp.Compiler.Service 34.1.1`

it works again.

baronfel:
Yes, this library needs to establish an upper bound on the version of FCS that it supports.  I try to adhere to SemVer when publishing FCS packages, but the FCS API is not stable and so any major version release could indicate non-backwards-compatible changes.

eiriktsarpalis:
> this library needs to establish an upper bound on the version of FCS that it supports.

We could perhaps pin to the major version, but maybe that's overly conservative? Not all breaking changes should necessarily impact FSharp.Formatting. As best practice I pin the FSCS version in the generate script dependencies.

baronfel:
I agree! I meant to pin at the major version. I'm using https://github.com/fsprojects/SyntacticVersioning in FCS to determine how the version number needs to increment, so I am reasonably confident that version bumps that are not major versions will be compatible in the way you describe.

eiriktsarpalis:
Fixed in 4.0.0-rc2

PR #513, opened 2020-04-24
Add upper border to FCS dependency

kMutagene:
See @baronfel 's suggestion at #511 .


eiriktsarpalis:
One thing to note is that, last time I checked, version ranges in paket.dependencies are not reflected in nuget packages, so this will have no effect on the generated artifact.

eiriktsarpalis:
I did in fact verify that this is the case by checking the [generated artifacts in AppVeyor](https://ci.appveyor.com/project/matthid/fsharp-formatting-32np3/builds/32407387/artifacts).

baronfel:
Good eye @eiriktsarpalis, it looks like the paket `fix-nuspecs` command, which is used to plug in to the dotnet sdk build pipeline, hasn't learned how to adjust version ranges that were generated by the SDK nuspec generate tasks.  It looks like the fix would need to be in this traversal function: https://github.com/fsprojects/Paket/blob/1ab743a5a928146059842ad7f35cc8501769ff80/src/Paket.Core/PublicAPI.fs#L878

kMutagene:
Yeah I missed that, i just expected the pack command to add the version constraints. @eiriktsarpalis  you seemed to have fixed #511 in #514 i will close this.

PR #517, opened 2020-05-27
Add support for customizing assigned CSS class

Krzysztof-Cieslak:
In my case I'd want to use standard hihglighting.js CSS classes rather than built-in ones - I'm using highlighting.JS for syntax highlighting of non-F# code snippets so I'd like to be consistent.


dsyme:
Looks good to me

PR #518, opened 2020-05-30
Support preview F# langauge features

cartermp:
I suppose the downside to these tests is that they may need to be updated over time as new preview features come in.

### DIFF
--- a/src/FSharp.Formatting.Common/YaafFSharpScripting.fs
+++ b/src/FSharp.Formatting.Common/YaafFSharpScripting.fs
@@ -201,6 +201,7 @@ module internal CompilerServiceExtensions =
             [| //yield "--debug:full"
                //yield "--define:DEBUG"
                //yield "--optimize-"
+               yield "--langversion:preview"
                yield "--nooptimizationdata"
                yield "--noframework"
 
--- a/tests/FSharp.CodeFormat.Tests/Tests.fs
+++ b/tests/FSharp.CodeFormat.Tests/Tests.fs
@@ -44,6 +44,65 @@ let ``Simple code snippet is formatted with tool tips``() =
     | _ -> false)
   |> shouldEqual true
 
+[<Test>]
+let ``Preview language feature from FSharp Core is supported``() =
+    let source = """
+let x = 12
+nameof x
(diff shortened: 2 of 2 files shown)


src/FSharp.Formatting.Common/YaafFSharpScripting.fs
```diff
@@ -201,6 +201,7 @@ module internal CompilerServiceExtensions =
             [| //yield "--debug:full"
                //yield "--define:DEBUG"
                //yield "--optimize-"
+               yield "--langversion:preview"
```
eiriktsarpalis, line 204: Is there any reason why users might need to opt out of that? Can't think of any myself.

tests/FSharp.CodeFormat.Tests/Tests.fs
```diff
@@ -44,6 +44,65 @@ let ``Simple code snippet is formatted with tool tips``() =
...
+    let snips, errors = agent.ParseSource("/somewhere/test.fsx", source.Trim())
+
+    errors |> shouldEqual [||]
+    snips |> containsSpan (function
+        | Token(_, "nameof", Some (ToolTipWithLiteral "val nameof : 'T -> string")) -> true
+        | _ -> false)
+    |> shouldEqual true
+
+[<Test>]
+let ``Preview language feature is supported``() =
```
eiriktsarpalis, line 62: I'll pedantically remark that this test will stop to serve its purpose once the particular feature set gets released. Perhaps add a comment noting that for posterity?

src/FSharp.Formatting.Common/YaafFSharpScripting.fs
cartermp: Possibly. You could imagine an fssnip showing off a feature in preview where we decide to change the syntax. That snippet would then break when fssnip updates to a later FCS. But in general I think the ability to show preview features outweighs the small likelihood that we would do something to a preview feature that led to big breaking changes).

Other than that, there's no real mechanical downside. The compiler used to generate data for formatting already has the preview codepaths, they're just not activated unless you flip this flag.

PR #527, opened 2020-07-06
Replace Razor by built in HTML generation + simple outer template

dsyme:
1. Remove Razor templating in `ApiDocs`

2. Provide and improve the `ApiDocsModel` (previously `GeneratorOutput`) so tools like Waypoint and MiniScaffold can continue to do their own templating over the integrated docs model

3. Make `ApiDocs.GenerateHtml` that does a reasonable job of generating quality HTML docs, currently essentially identical to previous default HTML doc generation.  There is no customization of these docs except via an outer template (with a simple substitution of {document} and properties), plus CSS styling.

In the future I'll aim to take the best from Waypoint to get searchable docs generated by default.


csharp/MarkdownSharpTests/App.config
```diff
@@ -28,22 +28,4 @@
   </log4net>
   <runtime>
     
-  <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
eiriktsarpalis, line 31: Can probably be deleted completely?

dsyme: yup

dsyme:
@eiriktsarpalis I decided to change this a lot.  I've decided that any dependency on any templating engine is a nightmare and basically unnecessary, also getting our layering wrong.

So, there will be no sophisticated templating customization in `FSharp.Formatting.ApiDocs` (previously called `FSharp.Formatting.MetadataFormat`). So there will be no dependency on *any* templating engine.  

Instead we will:

1. Continue to provide and improve the ApiDocs model so tools like Waypoint and MiniScaffold can continue to do high quality doc generation 

2. We will bake in a default generation of ApiDocs `ApiDocs.GenerateHtml` that does a reasonable job of generating quality HTML docs.  The implementation will be similar to Waypoint and I'll work with @Krzysztof-Cieslak on trying to get these aligned and getting searchable docs generated.  There will be **no customization** of these docs except via an outer template (with a simple substitution of `{document}` and properties), plus CSS styling.

PR #538, opened 2020-07-17
generate lunr index from 'fsdocs build'

dsyme:
1. build the Lunr `index.json` from every execution of `fsdocs build`

2. Make the search index entries available as part of the ApiDocs model

3. Add `ApiDocs` prefix to all types in `ApiDocsModel`

4. Remove `Details` from `ApiDocsModel`

5. Bump version to 6.0.1 since this is a binary change in the ApiDocsModel


dsyme:
@Krzysztof-Cieslak @baronfel @cartermp This now builds the Lunr search index on every execution of `fsdocs build`

The search index can also be built programatically using `ApiDocs.GenerateSearchIndexFromModel`

baronfel:
Taking a brief look at it, and one thing you pointed out in the docs repo issues was that RequireQualifiedAccess members didn't show up in the index when searched for like `List.<member>`. I didn't see any special handling here for RQA members and wasn't sure if you wanted to do that as part of this effort. It's _possible_ to recreate searches that do that based on a splitting + stemming pipeline in lunr, but it would be much easier to do simple prefix or suffix searches for the full member name.

dsyme:
>  I didn't see any special handling here for RQA members and wasn't sure if you wanted to do that as part of this effort. I

Yup this is done but for all `Module.foo` searches, not just RQA

dsyme:
@baronfel Search for things like `List.map` and any `Module.foo` are generated here: https://github.com/fsprojects/FSharp.Formatting/pull/538/files#diff-7f6959612189eb864b01a225e44e6684R91

PR #546, opened 2020-07-20
improve default styles and output

dsyme:
This adds more information upfront on members and improves styling for FSHarp.Formatting docs, reducing borders on tables etc.


dsyme:
@baronfel @Krzysztof-Cieslak Quite a lot of information is added to ApiDocsModel in this including base types and interfaces.

This information is now includes in the default formatting

PR #547, opened 2020-07-20
Format signature

Krzysztof-Cieslak:
This introduces nicely formatted signatures like ones used in the Ionide/VS4Mac/Rider

Current live version:
image

Updated version:
image

I believe with this change we don't need to list parameters and return type separately as it's clear from the signature.

### DIFF
--- a/docs/content/style.css
+++ b/docs/content/style.css
@@ -110,6 +110,15 @@ table.pre td.lines {
   width:30px;
 }
 
+div.tip .signature {
+ white-space: pre;
+ font-family: Monaco, Menlo, Consolas, 'Courier New', Courier, monospace;
+}
+
+div.tip .full-usage {
+  font-family: Monaco, Menlo, Consolas, 'Courier New', Courier, monospace;
+ }
+
 /*--------------------------------------------------------------------------
   Formatting for page & standard document content
 /*--------------------------------------------------------------------------*/
--- a/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
@@ -25,7 +25,7 @@ let renderEntities (entities: Choice<ApiDocType, ApiDocModule> list) =
           ]
         ]
         tbody [] [
-          for e in entities do 
+          for e in entities do
             tr [] [
                td [Class "type-name"] [
                  let nm = match e with Choice1Of2 t -> t.Name | Choice2Of2 m -> m.Name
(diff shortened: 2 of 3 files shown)


dsyme:
Hi, I only just saw this.  It will clash a bit with the changes I just pushed in sorry, but I think you'll be able to fit things together.

I like the signature for the hover tips, it looks great

dsyme:
@Krzysztof-Cieslak As a separate note, I'd like to go through and make all the type formatting generate HTML instead of strings

However this would mean [the currently internalised HtmlModel being used here](https://github.com/fsprojects/FSharp.Formatting/blob/master/src/FSharp.Formatting.Common/HtmlModel.fs) (note: ripped from Fornax) would be exposed as public.  

Since FSharp.Formatting is at the root of the universe I think this is ok?  But it could require translation from FSHarp.Formatting HTML --> Fornax HTML model

Do you think that HtmlModel should be in a single component all of its own, with a stable 1.0 version number?  I was loathe to take  a dependency on Fornax.Core v0.2 - for application code I would, but FSharp.Formatting is at the root of the entire universe and so I really want to avoid any significant dependencies for it.

baronfel:
The propagation of html rendering engines in the ecosystem is a big pain point IMO. This, fornax, giraffe, Feliz, falco all have different but similar implementations :-/

dsyme:
>The propagation of html rendering engines in the ecosystem is a big pain point IMO. This, fornax, giraffe, Feliz, falco all have different but similar implementations :-/

Right

I mean I've internalised it here so far, and it's currently not used in the ApiDocsModel.  The use in GenerateHtml is internal.

We could give the structuring of types in the APiDocsModel without it (as a separate data model).    I suppose that would be consistent with the current architecture and get the layering right

dsyme:
Let's discuss that here: https://github.com/fsprojects/FSharp.Formatting/issues/549

Krzysztof-Cieslak:
How about using Giraffe.ViewEngine? https://github.com/giraffe-fsharp/Giraffe.ViewEngine

It’s 0 dependency library that we have recently taken out of core Giraffe package to allow exactly this scenario - I was planning to remove HTML DSL from Fornax  Core and port Waypoint doc template to Giraffe.ViewEngine anyway

cartermp:
@dsyme it's not clear how to proceed with this one now that you've removed `SignatureTooltip`. Is there a replacement?

dsyme:
> @dsyme it's not clear how to proceed with this one now that you've removed SignatureTooltip. Is there a replacement?

I think the aim here is to now to put in a new, better SignatureTooltip. But the code will need to be rejigged to produce HTML rather than plain formatted text, I'll see if I can do that (since I'm the cause of most of the conflicts)

BTW we're now at the position that the HTML fragments being generated in ApiDocsModel are limited to:

1. F# types and signatures
2. The HTML for individual sections of XML comments (e.g. the HTML formatted for summary, remarks sections)
3. The HTML for individual sections of Markdown comments

This falls under category (1) so I think it's ok in terms of layering

cartermp:
Should this PR just be left open, or re-done?

dsyme:
>Should this PR just be left open, or re-done?

I will assess it, it's on my todo list.

dsyme:
Closing this old PR - @Krzysztof-Cieslak Apologies it never got merged.  We can look at merging into here of FCS

PR #568, opened 2020-08-04
Use GH actions for CI

cartermp:
let's see if this does stuff


cartermp:
This appears to work and is functionally equivalent to the travis and appveyor runs. The appveyor definition had something to do with NuGet artifacts but when I looked in the build log there were no artifacts produced and appveyor issued a warning about that. So I don't think anything regresses from switching to GH Actions.

I think the owner of the appveyor and travis account needs to sign in to unsubscribe or something, since they still show up in CI checks section but nothing happens. @dsyme ?

cartermp:
I was able to disable checks for appveyor/travis here: https://github.com/fsprojects/FSharp.Formatting/settings/branch_protection_rules/1463147

But they still show in the UI:

image

sergey-tihon:
I believe it was WebHooks. I just disabled them, so they should not be displayed anymore
image

cartermp:
In that case I think this is ready!

PR #615, opened 2020-10-14
Refactor the templating engine and the command tool cache et al.

### LINKED ISSUE #614: The command tool's cache is serialized with the BinaryFormatter
As seen in [`BuildCommand.fs`](https://github.com/fsprojects/FSharp.Formatting/blob/420e4a4eb3152fedd5ff647a68a73a5ea1fd79cd/src/FSharp.Formatting.CommandTool/BuildCommand.fs#L47), the project cracker cache the command tool stores in `.fsdocs/cache` is serialized using the [insecure and deprecated](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide) `BinaryFormatter` class.

To close this potential security hole, the command tool should use a safe serializer for its cache, such as the data contract serializer. Such change would break the existing cache files but the impact from this change is minimal.

teo-tsirpanis:
This PR:

* Fixes #614 by serializing the command tool's cache with the safer `DataContractSerializer` instead of the insecure `BinaryFormatter`. The XML output of the serializer is encided in binary to reduce its size.
* Rewrites the template engine in `Templating.fs` to use a more efficient `Span`-based single-pass algorithm.
* Optimizes the functions in the command tool's `Options.fs`, making them allocate less.
* Removes some unused source files and ignores built-time assets generated by FAKE, like `AssemblyInfo.?s` and `version.props`, the latter of which being now generated with LINQ-to-XML.
* _The rest of the changes will be described in code review comments._

---

As a consequence of the templating engine rewrite, the `FSharp.Formatting.Common` package multi-targets to .NET Standard 2.1 in addition to 2.0, to take advantage of the `StringBuilder.Append` overload that takes a `ReadOnlySpan` (in .NET Standard 2.0 the span's content will be first copied to a string). The command tool will automatically use the .NET Standard 2.1 version.

However, the other public packages -namely `FSharp.Formatting` and `FSharp.Formatting.Literate`- still target only .NET Standard 2.0 because of concerns about the packages' size being doubled. Should I multi-target them as well?


build.fsx
```diff
@@ -115,15 +114,15 @@ Target.create "NuGet" (fun _ ->
...
-    DotNet.exec id "tool" "uninstall --local FSharp.Formatting.CommandTool" |> ignore
-    // Use a local package store to avoid reuse of previous builds of the package with the same version
-    try 
-      Environment.setEnvironVar "NUGET_PACKAGES" (__SOURCE_DIRECTORY__ + "/.packages")
-      DotNet.exec id "tool" ("install --local --no-cache --version " + release.NugetVersion + " --add-source " + artifactsDir + " FSharp.Formatting.CommandTool")  |> ignore
-    finally
-      Environment.setEnvironVar "NUGET_PACKAGES" ""
-    DotNet.exec id "fsdocs" "build --strict --clean --property Configuration=Release" |> ignore
-    DotNet.exec id "tool" "uninstall --local FSharp.Formatting.CommandTool" |> ignore
+    // Τhe tool has been uninstalled when the
```
teo-tsirpanis, line 117: Using `--tool-path` in `dotnet tool install` is more reliable and not affected by any global installation of the command tool.

docs/_template.html
```diff
@@ -71,7 +71,7 @@ <h3 class="muted"><a href="{{fsdocs-collection-name-link}}">{{fsdocs-collection-
 
         <!-- BEGIN SEARCH BOX: this adds support for the search box -->
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/JavaScript-autoComplete/1.0.4/auto-complete.css" />
-        <script type="text/javascript">var fsdocs_search_baseurl = '{{root}}'</script>
+        <script type="text/javascript">var fsdocs_search_baseurl = '{{root}}';</script>
```
teo-tsirpanis, line 74: The semicolon fixes a syntax highlighter glitch in VSCode.

src/FSharp.Formatting.Common/FSharp.Formatting.Common.fsproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0; netstandard2.1</TargetFrameworks>
```
dsyme, line 4: Why both?

src/FSharp.Formatting.Common/Templating.fs
```diff
@@ -1,25 +1,19 @@
...
+open System.Text
 
-#if DEBUG
+[<Struct>]
 type ParamKey = ParamKey of string
-#else
-/// An abbreviation for 'string' representing a strong name for a parameter key
-type ParamKey = string
-#endif
+with
```
dsyme, line 10: This "with" is not needed

dsyme: (And will not compile in release mode??)

dsyme:
> However, the other public packages -namely FSharp.Formatting and FSharp.Formatting.Literate- still target only .NET Standard 2.0 because of concerns about the packages' size being doubled. Should I multi-target them as well?

Yes just multi-target them, or move them all to .NET Standard 2.1 (probably ok?)

dsyme:
Thanks, this looks great.  Just move everything to netstandard 2.1 I think?

PR #618, opened 2020-10-18
Refactor the project cracker into a new file et al.

### LINKED ISSUE #617: The command tool does not correctly parse custom substitution parameters
If we run `dotnet fsdocs build --parameters foo bar yoshi sbubby`, the command tool prints

```
Substitutions/parameters:
  foo --> bar
  bar --> yoshi
  yoshi --> sbubby
```

while the correct answer should have been

```
Substitutions/parameters:
  foo --> bar
  yoshi --> sbubby
```

Moreover, if we pass an odd number of arguments to `--parameters` the tool should fail but it doesn't.

---

The problem is that the function that parses the arguments uses `Seq.pairwise` which groups the items of a sequence in _overlapping_ pairs.

teo-tsirpanis:
This PR:

* Reduces the size of `BuildCommand.fs` by moving the project cracker in a file called `ProjectCracker.fs`.
* Fixes #617 by correctly grouping the arguments for the command's tool `--parameters` option and document how to use that command.
* Makes the command tool stop reporting non-existent substitution parameters (they used to be reported as `fsdocs-repository-link --> {{fsdocs-repository-link}}`).
* Removes the command tool's dependency to Newtonsoft.Json in favor of `System.Text.Json` in the documentation search index serializer.
* Updates the packages, decluttering `paket.lock`.
* Removes unnecessary code and optimizes performance in  acouple of other cases.


src/FSharp.Formatting.CommandTool/BuildCommand.fs
```diff
@@ -714,22 +309,20 @@ type CoreBuildOptions(watch) =
...
-                    printfn "Error : \n%O" ex
-                    if this.strict then exit 1
-                    false
-                | ex ->
-                    Log.errorf "received exception :\n %A" ex
-                    printfn "Error : \n%O" ex
-                    if this.strict then exit 1
-                    false
+            with ex ->
+                Log.errorf "received exception :\n %A" ex
```
wallymathieu, line 313: Doesn't seem to be any different logic related to aggregate exceptions 👍

src/FSharp.Formatting.CommandTool/Options.fs
```diff
@@ -9,10 +9,20 @@ module Common =
...
+            if enumerator.MoveNext() then
+                let second = enumerator.Current
+                yield first, second
+    }
+
     let evalPairwiseStrings a =
         match Seq.tryExactlyOne a with
         | Some "" -> None
-        | _ -> a |> Seq.pairwise |> List.ofSeq |> Some
+        | _ -> a |> pairs |> List.ofSeq |> Some
```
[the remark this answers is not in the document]
dsyme: See https://stackoverflow.com/questions/4126351/how-to-get-pairs-of-consecutive-values-from-f-seq

dsyme (approved):

src/FSharp.Formatting.CommandTool/ProjectCracker.fs
```diff
@@ -0,0 +1,415 @@
...
+            "PackageLicenseExpression"
+            "PackageTags"
+            "Copyright"
+            "PackageVersion"
+            "PackageIconUrl"
+            //Removed because this is typically a multi-line string and dotnet-proj-info can't handle this
+            //"PackageReleaseNotes"
+            "RepositoryCommit"
+        ]
+        let gp() = Inspect.getProperties ("TargetPath" :: additionalInfo)
```
dsyme, line 169: This abbreviation is too short, just use a proper name please

teo-tsirpanis: It was already named like that; I left it as-is.

dsyme: ah ok :-)

Issue #611, opened 2020-09-25
Ability to output a raw string from an FSI evalutation instead of having it wrapped in code blocks


HLWeil:
Hey, is it possible to somehow use a custom `fsiEvaluator` when running `dotnet fsdocs build`?

dsyme:
> Hey, is it possible to somehow use a custom fsiEvaluator when running dotnet fsdocs build?

No.  This kind of parameterization sits very awkwardly with a command line tool

What specifically are you looking for ?  I never fully understood the use cases for custom evaluators, given the complexity they bring

HLWeil:
Thanks for your answer.

I'm looking for embedding Plotly charts like we did [here](https://csbiology.github.io/FSharp.Stats/#Regression).

HLWeil:
Sry my description was very unspecific. 
My problem is, that even when I try a workaround like piping the chart into a function which returns a finished html string, this string then gets inserted intro `<pre>`  tags, preventing the chart to be rendered. In our own build script, we used an FSI evaluator which did not create this tag when given a chart. But using the new command line tools seems much more lightweight and easily manageable so we would like to migrate.

I understand that using the custom evaluator with the tools is not doable. So the follow up question would be: Is it possible to prevent the standard evaluator from packing the evaluated result into `<pre>` tags?

dsyme:
OK I see. Yes we can address this somehow

However I'd really appreciate step by step instructions about which charting library etc. Exact step by step please, including actual and expected results

HLWeil:
## Reproduction
I created a repo for illucidating my problem: https://github.com/HLWeil/FsDocsEvaluatedCharts

The repro steps are described there. The created `output/ChartTest.html` is the file of interest. In this file three ways to embed a chart are displayed:

1.  `Include Chart directly` (Using a custom FsiEvaluator; Undoable)

2. `Include Evaluated HTML string` (Using an additional function)

3.  `HTML String` (Include the chart html **by hand**; Just for showcase purposes)

## Desired result

The chart html gets correctly formatted not only in the third, but also in the second case.

## Actual result

The chart html does not get formatted correctly but as a code snippet. This is because the standard FsiEvaluator inserts the evaluated result into `<code>` and `<pre>` tags.

## Possible solution

Maybe it is possible to offer a `literate command` which does not encapsulate the evaluated result into tags. This could e.g. be called `(***include-inline***)`.

adz:
[...]

PR #624, opened 2020-12-20
Add include-it-raw literate command

HLWeil:
This pull request addresses issue #611. 

The idea is to have a way to include `results of snippets` not formatted as fsi output but just as `plain text`. This allows for formatting them as `html`, e.g. as `charts`, as can be seen [here](https://fslab.org/FSharp.Stats/Covariance.html). In the example, this was done by using a `custom evalutator`, which is not feasible with the `fsdocs tool`.

Therefore I added a new literate command `include-it-raw`, which evaluates the snippet, but then inserts is as plain text. As a small example I used this to embed a `html table` from a jagged array:

IncludeItRaw

The code in the `Evaluator.fs` got a little bit messy, as `string`s are formatted surrounded by `"`s, which show up in the final document and might even hinder the correct html formatting. I remove them in the `Evaluator.Format` method but maybe you have a better place or idea?

### DIFF
--- a/docs/literate.fsx
+++ b/docs/literate.fsx
@@ -64,6 +64,7 @@ The F# script files is processed as follows:
 |   `(*** include-fsi-output ***)`       | The F# Interactive output of the preceeding snippet   |
 |   `(*** include-fsi-merged-output ***)`       | The merge of console output and F# Interactive output of the preceeding snippet   |
 |   `(*** include-it ***)`       | The formatted result of the preceeding snippet |
+|   `(*** include-it-raw ***)`       | The unformatted result of the preceeding snippet |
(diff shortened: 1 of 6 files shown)


cartermp:
@HLWeil could you look into adding a test here? https://github.com/fsprojects/FSharp.Formatting/blob/master/tests/FSharp.Literate.Tests/EvalTests.fs

HLWeil:
Sure thing! Something like this?

cartermp (approved):
Thanks, this is a good addition. I take it this lets you also output the HTML string for a chart yes?

cartermp:
Will merge, as this is a good change that does what you're after here.

Issue #627, opened 2021-01-06
Live refresh/Hot reload


kMutagene:
Currently, when developing docs via `fsdocs watch`, one has to manually reload the browser page after the new files are generated. As this project uses Suave, it should be possible to mirror the changes of [this PR](https://github.com/ionide/Fornax/pull/29) to fornax, which implemented live refresh there.

dsyme:
Sounds good

Kevin, I think I should make you Co admin on this repot and nuget (I'm on vacation and haven't republished the nuget yet)

dsyme:
Kevin, what's your nuget id please?

kMutagene:
I'll be glad to help where I can, sorry to have bothered you during vacation, enjoy the rest of it ;). My nuget profile is here: https://www.nuget.org/profiles/Mutagene.

dsyme:
@kMutagene I've sent the request on github

kMutagene:
@dsyme thank you! I was just writing the release docs for the next package (feels a little strange to write those about commits that are not my own) and wondering about #621 and how that commit affects the version number of the next package, will it be 8.0 or 7.3? When that is sorted i will release the package today

dsyme:
@kMutagene When in doubt bump the major version :-)

kMutagene:
Closed as of #629

PR #629, opened 2021-01-11
Add hot reload for the watch command (#627)

kMutagene:
This is a WIP PR trying to implement hot reload for the `fsdocs watch` command as discussed in #627.

The server side part is done, and the cycle works when adding the following script to the _template.html files:

```javascript
var wsUri = "ws://localhost:8901/websocket";
function init()
{
    websocket = new WebSocket(wsUri);
    console.log('connected')
    websocket.onclose = function(evt) { onClose(evt) };
}
function onClose(evt)
{
    console.log('closing');
    websocket.close();
    document.location.reload();
}
window.addEventListener("load", init, false);
```

But i am not sure what the best way to handle injecting this code is, as it has to be removed when building the actual docs, or the page will be loked in a constant reload cycle. My idea so far would be, when running the watch command, copy the template files, inject the code, and generate files based on that. This would also need an additional handling of updating these copied template files. I am not familiar enough with the code base yet to see what the best approach here is, and where best to intoduce these changes in the code, so feedback is welcome!


src/FSharp.Formatting.CommandTool/BuildCommand.fs
```diff
@@ -643,33 +651,43 @@ type CoreBuildOptions(watch) =
...
                             if runGeneratePhase2() then
-                                regenerateSearchIndex()) }))
+                                regenerateSearchIndex())
+                    }
+                    |> Async.RunSynchronously
+                    Serve.signalHotReload <- true
+                )
 
+            // handler for signalling to refresh the page to the client
+            let hotReloadHandler _ = Serve.signalHotReload <- true
```
dsyme, line 685: Does this `hotReloadHandler ` get used anywhere?

dsyme:
> But i am not sure what the best way to handle injecting this code is, as it has to be removed when building the actual docs, or the page will be loked in a constant reload cycle. My idea so far would be, when running the watch command, copy the template files, inject the code, and generate files based on that. This would also need an additional handling of updating these copied template files. I am not familiar enough with the code base yet to see what the best approach here is, and where best to intoduce these changes in the code, so feedback is welcome!

I'm not sure of the best approach here - perhaps a template substitution that's only activated when building `watch` docs?

src/FSharp.Formatting.CommandTool/BuildCommand.fs
kMutagene: nope, that is indeed a leftover from a previous test, will remove

kMutagene:
> I'm not sure of the best approach here - perhaps a template substitution that's only activated when building watch docs?

That worked well!

dsyme:
@kMutagene This looks ready?

kMutagene:
@dsyme yeah i was just not sure if i should merge my own PR or wait for a review ;)

dsyme:
> @dsyme yeah i was just not sure if i should merge my own PR or wait for a review ;)

Feel free to ask for a review bia @ to myself and other maintainters.  Then if you don't get one after a while you can self-merge :-)

docs/styling.md
```diff
@@ -104,6 +104,12 @@ If you write a new theme by CSS styling please contribute it back to FSharp.Form
 You can do advanced styling by creating a new template.  Add a file `docs/_template.html`, likely starting
 with the existing default template.
 
+to enable hot reload during development with `fsdocs watch` in a custom `_template.html` file, make sure to add the following line to your `<head>` tag:
```
dsyme, line 107: Capital "T" at start of line.

Also there is a table of substitutions in content.md I believe, this should be added there

src/FSharp.Formatting.CommandTool/BuildCommand.fs
```diff
@@ -643,33 +683,40 @@ type CoreBuildOptions(watch) =
...
                         do! Async.Sleep(300)
                         lock monitor (fun () ->
                         docsQueued <- false
                         if runDocContentPhase1() then
                             if runDocContentPhase2() then
                                 regenerateSearchIndex()
-                        ) }) )
+                        )
+                    }
+                    |> Async.RunSynchronously
```
dsyme, line 695: This has changed from Start to RunSynchronously.  If it's RunSynchronously then no async is needed?

src/FSharp.Formatting.CommandTool/BuildCommand.fs
```diff
@@ -643,33 +683,40 @@ type CoreBuildOptions(watch) =
...
                         do! Async.Sleep(300)
                         lock monitor (fun () ->
                         generateQueued <- false
                         if runGeneratePhase1() then
                             if runGeneratePhase2() then
-                                regenerateSearchIndex()) }))
-
+                                regenerateSearchIndex())
+                    }
+                    |> Async.RunSynchronously
```
dsyme, line 712: Likewise here....

dsyme:
On Start v. RunSynchronously - was this fixing a bug?

src/FSharp.Formatting.CommandTool/BuildCommand.fs
kMutagene: You are right, this is unnecessary. I was kind of thrown off by the 'fire and forget' nature of Async.Start (via [this](https://docs.microsoft.com/de-de/dotnet/fsharp/tutorials/asynchronous-and-concurrent-programming/async#asyncstart)) and ditched it without second thought thinking i have to wait until the computation is finished. I missed that the hot reload trigger does not have to happen after the whole block but just after the last function call inside it, so i moved it inside the block now using the original Async.start version again.

PR #634, opened 2021-01-17
Prevent CLI parameters from being discarded

### LINKED ISSUE #633: Regression: CLI "--parameters" are ignored
Found a regression in v8 (n.b. v7.2.7 _does not_ exhibit this behavior).

#### Expected Behavior

Values passed via the `--parameters` argument override values set in `.fsproj` (which, in turn override hard-coded default values).

#### Actual Behavior

Values passed via the `--parameters` argument are ignored.

#### Steps to reproduce

1. invoke `dotnet fsdocs watch --projects <path/to/fsproj> --parameters fsdocs-navbar-position fixed-left`
2. observe in the console output that _mulitple_ values are listed for a given parameter (the default, from `.fsproj` (if given), from CLI)
3. observe in the generated output that the value from the `.fsproj` file is used (if present) _or_ the default value is used.

#### Environment

+ .NET SDK version: 5.0.102
+ Operating System: Pop!_OS 20.10 (Ubuntu-derivative)

pblasucci:
Fixes #633 

This took a little while to hunt down. But, basically, the logic is:

1. Accumulate parameters from multiple sources
2. Rely on FSharp.Core's `dict` and `readOnlyDict` to de-duplicate values

The limitation of the above approach is that it is _order dependent_. This isn't bad, per se. But it means step one above needs to be handled carefully. So, this patch just changes the order in which things are accumulated (putting the `userParameters` at the _end_ of the list).

### DIFF
--- a/src/FSharp.Formatting.CommandTool/ProjectCracker.fs
+++ b/src/FSharp.Formatting.CommandTool/ProjectCracker.fs
@@ -402,7 +402,7 @@ module Crack =
         let parametersForProjectInfo (info: CrackedProjectInfo) =
             let projectUrl = info.PackageProjectUrl |> Option.map ensureTrailingSlash |> Option.defaultValue root
             let repoUrl = info.RepositoryUrl |> Option.map ensureTrailingSlash
-            userParameters @ List.choose id [
+            List.choose id [
                 param None ParamKeys.``root`` (Some root)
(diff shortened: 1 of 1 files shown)


kMutagene:
Tested and it works as you say, great catch, thanks!

PR #640, opened 2021-02-11
upgrade to FCS 39.0.0 and change --property to --properties

dsyme:
upgrade to FCS 39.0.0 

change --property to --properties

minor doc fixes


dsyme:
@kMutagene Note this is a breking change, fsdocs `--property` argument is now called `--properties`

We could bump to 9.0 I guess

PR #645, opened 2021-02-15
use recommended mathjax cdn

nhirschey:
- fsdocs uses cdn.mathjax.org, which is deprecated: https://www.mathjax.org/cdn-shutting-down/
- The recommended cdn is cdnjs.cloudflare.com: http://docs.mathjax.org/en/v2.7-latest/start.html#using-a-content-delivery-network-cdn. Fsdocs temporarily used the recommended cdn, but it was commented out and replaced in this commit from July 2020: https://github.com/fsprojects/FSharp.Formatting/commit/f3ea7de4244ef935b5a95fcea6b4c2353341aece

Is there a reason we're not using the recommended cdn? I had issues where mathjax would randomly work/not work on a page in fsdocs watch mode, and switching to the recommended cdn at cloudfare resolved the issue.


dsyme (approved):

dsyme:
> Is there a reason we're not using the recommended cdn? I had issues where mathjax would randomly work/not work on a page in fsdocs watch mode, and switching to the recommended cdn at cloudfare resolved the issue.

No reason. Thanks!

PR #651, opened 2021-03-24
Fix default styles to show examples and strings more clearly

dsyme:
There are some glitches in the CSS defaults:

1. Examples not shown in clearly contrasted font:

image

2. Strings not shown in clearly visible font:

image


dsyme:
Actually this is just for the light style of DiffSharp repo

Issue #650, opened 2021-03-18
fsdocs watch failure in WSL


catfly1:
### Problem
`fsdocs watch` on WSL does not work as browsers are not installed inside WSL (generally). This can be an issue for users who are on [VS Code Remote](https://code.visualstudio.com/docs/remote/remote-overview).

### Isolated test case
 Works on Windows FSI, fails on WSL fsi:

`System.Diagnostics.Process.Start(new System.Diagnostics.ProcessStartInfo("http://test", UseShellExecute = true)) |> ignore`

### Stack trace
Stack trace caused [by this line](https://github.com/fsprojects/FSharp.Formatting/blob/master/src/FSharp.Formatting.CommandTool/BuildCommand.fs#L816):
```
launching browser window to open http://localhost:8901/
[20:05:17 INF] Smooth! Suave listener started in 15.012ms with binding 127.0.0.1:8901
Unhandled exception. System.ComponentModel.Win32Exception (2): No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
   at FSharp.Formatting.CommandTool.CoreBuildOptions.Execute() in /home/runner/work/FSharp.Formatting/FSharp.Formatting/src/FSharp.Formatting.CommandTool/BuildCommand.fs:line 816
   at FSharp.Formatting.CommandTool.Main.main@10-1.Invoke(WatchCommand opts) in /home/runner/work/FSharp.Formatting/FSharp.Formatting/src/FSharp.Formatting.CommandTool/Program.fs:line 10
   at CommandLine.ParserResultExtensions.MapResult[T1,T2,TResult](ParserResult`1 result, Func`2 parsedFunc1, Func`2 parsedFunc2, Func`2 notParsedFunc)
[...]

PR #654, opened 2021-03-24
Make Process.Start failure non fatal for browser start

catfly1:
Handle exceptions: https://github.com/fsprojects/FSharp.Formatting/issues/650

I couldn't find a good spot to put CLI related integration test cases into the project.


dsyme:
Thanks!  It's ok, no test needed for this one

PR #681, opened 2021-06-12
Improve github and signature links

gbaydin:
This improves the api docs github source and signature links (copy xml/md to clipboard) with tooltips and updated icons.


gbaydin:
Can we have a new nuget release? https://www.nuget.org/packages/FSharp.Formatting/

PR #682, opened 2021-06-29
Bump to FCS 40

baronfel:
This updates FCS to version 40, partially because we in FSAC use FSharp.Formatting internally so we need this to be upgraded before we can move forward :D

I think the only really controversial thing here is the change from netcoreapp3.1 to net5.0, because the Ionide/proj-info dependency requires a version of MSBuild that no longer seems to provide a netcoreapp3.1 binary.  The tests all seem to be passing after this change as well.


dsyme (approved):

dsyme:
@baronfel Thank you for this. The changes are fine

I will make you co-admin in this repo if you like?

baronfel:
That would be fine with me! I promise to use my powers only for good. My main use case is probably going to be FCS bumps to enable FSAC, so I don't plan on going crazy with foundational redesigns or anything :)

Issue #686, opened 2021-07-21
The new default css based on diffsharp removed bottom margins from code output.


nhirschey:
See below screenshot from FSharp.Data. The orange is the margin. There is a bottom margin for code, but the second code block (which is showing the csv file) is missing the orange bottom margin.

![image](https://user-images.githubusercontent.com/5226115/126525769-00969454-9b25-4426-95b8-462bbd68e746.png)

A simple fix is to change this  [line](https://github.com/fsprojects/FSharp.Formatting/blob/11aa8d0c9a1163e038ec545298fe5d1184940b6e/docs/content/fsdocs-default.css#L328) that was commented out in this [pull request](https://github.com/fsprojects/FSharp.Formatting/pull/664) to `margin-bottom: 1em;`. That will give the same margin for both code and code output.

With the change, see the bottom computed margin for code:
<img width="492" alt="code" src="https://user-images.githubusercontent.com/5226115/126528458-042c7e20-56ed-499d-ab47-1664ffb564be.png">

And for output:
<img width="668" alt="output" src="https://user-images.githubusercontent.com/5226115/126528474-cbbfab05-91d7-469c-a2b9-4ffb30f70962.png">

PR #687, opened 2021-07-21
add bottom margin to included code output

nhirschey:
Addresses issue raised here https://github.com/fsprojects/FSharp.Formatting/issues/686

### DIFF
--- a/docs/content/fsdocs-default.css
+++ b/docs/content/fsdocs-default.css
@@ -325,7 +325,7 @@ body {
     white-space: pre;
     font: 0.85rem 'Roboto Mono', monospace;
     width: 90%;
-    /* margin: 10px 20px 20px 20px; */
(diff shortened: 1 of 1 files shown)


dsyme:
Thanks!

dsyme:
@nhirschey I think @gbaydin has recently made some more improvements to DiffSharp CSS, we should trial them for incorporation back into here

gbaydin:
I can send a quick PR with the latest one I'm quite happy with.

dsyme:
Yes please do if you have time, @nhirschey if you could review, trial and compare that would be great

@gbaydin please copy back any recent changes line by line rather than wholesale, the CSS wasn't completely aligned

nhirschey:
Sure, I'm happy to review and compare. I like the diffsharp style. So far this is the only style "regression" that I noticed.

PR #688, opened 2021-07-22
Updated DiffSharp css

gbaydin:
@dsyme this is the latest DiffSharp css as we mentioned in #687. I'm not sure how I could do the line by line copying. This PR is based on the latest master. I'm hoping it can give an indication of what needs to be merged.

### DIFF
--- a/docs/content/fsdocs-default.css
+++ b/docs/content/fsdocs-default.css
@@ -1,13 +1,11 @@
-/* @import url(https://fonts.googleapis.com/css?family=Droid+Sans|Droid+Sans+Mono|Open+Sans:400,600,700); */
-/* @import url('https://fonts.googleapis.com/css2?family=Hind+Vadodara&display=swap'); */
+/* CSS file from the fsprojects/FSharp.Formatting project modified by Atilim Gunes Baydin <gunes@robots.ox.ac.uk> for the DiffSharp project documentation, July 2021 */
+
 @import url('https://fonts.googleapis.com/css2?family=Hind+Vadodara&family=Roboto+Mono&display=swap');
 /*--------------------------------------------------------------------------
   Formatting for page & standard document content
 /*--------------------------------------------------------------------------*/
 
 body {
-    /* font-family: 'Open Sans', serif; */
-    /* font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif; */
     font-family: 'Hind Vadodara', sans-serif;
     /*    padding-top: 0px;
     padding-bottom: 40px;
@@ -50,6 +48,7 @@ body {
     font-size: 2rem;
     letter-spacing: 1.78px;
     line-height: 2.5rem;
+    text-transform: uppercase;    
     font-weight: 400;
 }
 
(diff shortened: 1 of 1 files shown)


dsyme:
Cool looking at the diff there are a few lines we can integrate here, I'll edit it down to be just those

docs/content/fsdocs-default.css
```diff
@@ -586,3 +592,41 @@ div.fsdocs-tip {
         color: #d1d1d1;
         font: 0.85rem 'Roboto Mono', monospace;
     }
+
+.button {
```
dsyme, line 596: Do you remember what these changes are for?  In general the fsdocs CSS doesn't give styline apart from specific fsdocs class and known element names

docs/content/fsdocs-default.css
```diff
@@ -1,13 +1,11 @@
-/* @import url(https://fonts.googleapis.com/css?family=Droid+Sans|Droid+Sans+Mono|Open+Sans:400,600,700); */
-/* @import url('https://fonts.googleapis.com/css2?family=Hind+Vadodara&display=swap'); */
+/* CSS file from the fsprojects/FSharp.Formatting project modified by Atilim Gunes Baydin <gunes@robots.ox.ac.uk> for the DiffSharp project documentation, July 2021 */
```
dsyme: Because these are a small subset of contributions back to fsdocs I removed this if that's ok, will metion in the release notes

dsyme:
@gbaydin I direct committed to your fix3 patch branch trimming out some of the diff

docs/content/fsdocs-default.css
```diff
@@ -323,36 +319,45 @@ body {
     border-top: 0px solid #e3e3e3;
     border-collapse: separate;
     white-space: pre;
-    font: 0.85rem 'Roboto Mono', monospace;
-    width: 90%;
-    margin-bottom: 1em;
+    font: 0.86rem 'Roboto Mono', monospace;
+    width: 100%;
+    margin: 10px 0px 20px 0px;
```
dsyme, line 324: @nhirschey You'll want to check this change

docs/content/fsdocs-default.css
```diff
@@ -540,6 +543,8 @@ body {
 /* comment */
 #fsdocs-content span.c {
     color: #808080;
+    font-weight: 400;
+    font-style: italic;
```
dsyme, line 547: @gbaydin Any recollection what this was changing? thanks

gbaydin: That's about making the code comments a bit lightweight and italic. Somewhat influenced by tensorflow docs style.

docs/content/fsdocs-default.css
gbaydin: I think this does fix the margings issue, but would be good to check if it removes the need for `margin-bottom: 1em;`

docs/content/fsdocs-default.css
gbaydin: That's fine

docs/content/fsdocs-default.css
gbaydin: This is for the install button we have on the home page. Can be removed if it doesn't apply to typical fsdocs usage.
image

dsyme: Thanks, yes, removed that bit.

dsyme:
@nhirschey I'll merge, release 11.4 and you can check

dsyme:
@nhirschey 11.3.0 doesn't have these changes, 11.4.0 has them.

Issue #672, opened 2021-04-28
Menu is not scrollable on mobile


nhirschey:
Following up on comment here https://github.com/fsprojects/FSharp.Data/pull/1370#issuecomment-820428350

Summary: if you’re on a mobile browser then clicking the sidebar/hamburger on the current template fills the browser window with the sidebar, and that sidebar menu that pops out is not scrollable. So it’s impossible (at least on iOS) to scroll to and click on links at the bottom of the side bar. You’re restricted to the first few items on the sidebar. In the screenshot in the above linked FSharp.Data comment, I can’t access anything below XML Provider.

dsyme:
Thanks for the fix!

dsyme:
@nhirschey Scrollable navbar looks a bit off for desktop, see

https://fsprojects.github.io/FSharp.Formatting/

![image](https://user-images.githubusercontent.com/7204669/126667262-cd79bb57-29e2-48df-bf3b-67f56baae7dd.png)

nhirschey:
Ok, I will see if I can find a way to expand it so that it fills the vertical real estate on desktop. In your screenshot the scrollbar isn’t needed.

nhirschey:
Pull request #689 is the most minimal change to the template that
1. Has no navbar scrollbar on desktop when there is sufficient vertical space to display all navbar items.
2. Makes the navbar usable on mobile browsers (by adding a scrollbar, see https://github.com/fsprojects/FSharp.Data/pull/1370#issuecomment-820428350 for a description of the problem).

That said, the current navbar in the FSharp.Formatting template could probably be simplified. I will experiment as I have time.

dsyme:
Thank you! that's great

PR #689, opened 2021-07-22
fix navbar scroll on desktop

nhirschey:
This fixes the desktop navbar scroll issue mentioned here https://github.com/fsprojects/FSharp.Formatting/issues/672#issuecomment-885012450

In the below gif, I first show the desktop experience with the navbar scroll only showing up when the browser window is shrunk vertically. Then I show the experience on mobile by shrinking horizontal space; if you click on the navbar button the scrollbar is visible.

fixed-scroll

### DIFF
--- a/docs/_template.html
+++ b/docs/_template.html
@@ -29,11 +29,11 @@
 </head>
 
 <body>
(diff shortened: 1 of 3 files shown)


dsyme (approved):

dsyme:
Thanks, I checked and it works well now

Issue #692, opened 2021-07-29
Cache needs removing if <PackageProjectUrl> is edited


davedawkins:
To reproduce:
1. Run `fsdocs build`
2. Note value of `root` in output log under `Substitutions/parameters`
3. Edit value of `<PackageProjectUrl>`
4. Re-run `fsdocs build`
5. Note value of `root` hasn't changed (it should have value used in step 3 though)
6. Remove `.fsdocs/cache`
7. Re-run `fsdocs build`
8. Note value 'root' now reflects value of `<PackageProjectUrl>`

Eg. Edited .fsproj file
```
    <PackageProjectUrl>https://pkgurl/abc1</PackageProjectUrl>
```

Output from `fsdocs build`
```
david@davebook Sutil % dotnet fsdocs build --output public/docs  |more
restored project state from '.fsdocs/cache'

Inputs for API Docs:
    /Users/david/projects/Sutil/src/Sutil/bin/Debug/netstandard2.0/Sutil.dll

Substitutions/parameters:
  root --> https://pkgurl/abc/
```

davedawkins:
I'll attempt a PR for this

davedawkins:
PR incoming...

dsyme:
@davedawkins Did you send a PR for this?

davedawkins:
I appear to have committed to my own fork and then not progressed to sending a PR. Investigating...

davedawkins:
OK I worked out what happened. I did submit a PR, then realized it could be replaced with a doc update, and we closed the PR. See https://github.com/fsprojects/FSharp.Formatting/pull/698

dsyme:
Cool, thank you for checking!

PR #698, opened 2021-07-30
Fix for #692

davedawkins:
Use timestamps from project files to validate cache.

Ideally, timestamps would be part of the cache key, but we'd have to call the project cracker to find the input files, and this would defeat the purpose of the cache. Instead, we pull them out of the cached value and recalculate.

All tests pass as run from `build.sh`. Also tested against my own project


davedawkins:
*Now* I understand that I have to build the DLL (and therefore produce a new API docs XML file) before any edits to the API docs will be observed. That means this PR can probably be replaced by a docs update.

dsyme:
>  That means this PR can be replaced by a docs update.

OK, I will close this PR :)

thanks for the doc updates!

PR #701, opened 2021-08-17
add blockquote style

nhirschey:
Markdown defines blockquote using ">" but the current css does not style these elements. This pull request adds the blockquote style used by the markdown inventor Jon Gruber at his site [daringfireball.net](https://www.daringfireball.net).

Example script:
```fsharp
(**
This is some text.

> This is some text in a block quote.
>
> Another paragraph with the same text.
>    
> 1. Item 1
> 1. Item 2
>  
> ```fsharp
> let y = 4
> ```

A later paragraph
*)

let x = 4.4

(*** include-fsi-output ***)
```

### This is the result after the pull request:
image

### This is using the current css style:
image

### DIFF
--- a/docs/content/fsdocs-default.css
+++ b/docs/content/fsdocs-default.css
@@ -10,6 +10,17 @@ body {
 */
 }
 
+blockquote {
+    margin: 0 1em 0 0.25em;
+    margin-top: 0px;
+    margin-right: 1em;
+    margin-bottom: 0px;
+    margin-left: 0.25em;
+    padding: 0 .75em 0 1em;
+    border-left: 1px solid #777;
+    border-right: 0px solid #777;
+}
+
 /* Format the heading - nicer spacing etc. */
 .masthead {
     overflow: hidden;


dsyme:
Thanks!

dsyme:
So in general we label all outputs with `fsdocs-foobar` styles and style those.  Should we be doing that here?

My thinking was that we might want to be able to embed fsdocs outputs into other stylings

dsyme:
@nhirschey Would you like to be co-owner on this with me, so you can release if necessary, e.g. by this kind of auto-commit?

https://github.com/fsprojects/FSharp.Formatting/commit/23727d09d7b57ac3cf14f3c14cf436a673c1f714#diff-68d24fa81558aae3d8c59e2aa57a4fa719ea3b04d7fa14beff45f16f00858f50


(Or if you include the RELEASE_NOTES.md update in the PR with version bump the pipeline will auto-release on merge to main)

nhirschey:
> Would you like to be co-owner on this with me, so you can release if necessary, e.g. by this kind of auto-commit?

Sure, if it helps. I am an enthusiastic user of fsdocs and I understand the need for help. But you should know that I probably don't have the skillset or time to help if major architectural changes like the ones you made the past year were needed.


Re: autocommit, understood, I am familiar with your auto-deploy via release notes workflow. 

> So in general we label all outputs with fsdocs-foobar styles and style those. Should we be doing that here?

Got it, I can address that.

PR #704, opened 2021-09-08
Allow user-set ids for xmldoc example nodes

baronfel:
Based on [this discussion](https://github.com/dotnet/fsharp/pull/12130#discussion_r704477523) it seems like we should support user-defined ids for examples (for example to allow deep-linking to a specific example).

### DIFF
--- a/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
@@ -215,7 +215,7 @@ type HtmlRender(model: ApiDocModel) =
 
                   for e in m.Comment.Examples do
                       h5 [Class "fsdocs-example-header"] [!! "Example"]
-                      p [Class "fsdocs-example"] [embed e]
+                      p [Class "fsdocs-example"; if e.Id.IsSome then Id e.Id.Value ] [embed e]
 
                   //if m.IsObsolete then
                   //    obsoleteMessage m.ObsoleteMessage
--- a/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
@@ -264,7 +264,7 @@ type ApiDocMember (displayName: string, attributes: ApiDocAttribute list, entity
          comment: ApiDocComment, symbol: FSharpSymbol, warn) =
 
   let (ApiDocMemberDetails(usageHtml, paramTypes, returnType, modifiers, typars, baseType, location, compiledName)) = details
-
+  let m = defaultArg symbol.DeclarationLocation range0
   // merge the parameter docs and parameter types
   let parameters =
       let paramTypes =
@@ -276,7 +276,7 @@ type ApiDocMember (displayName: string, attributes: ApiDocAttribute list, entity
                   (psym, pnm, _pnameText, _pty) ]
       let tnames = Set.ofList [ for (_psym, pnm, _pnameText, _pty) in paramTypes -> pnm ]
       let tdocs = Map.ofList [ for pnm, doc in comment.Parameters -> Some pnm, doc ]
-      let m = defaultArg symbol.DeclarationLocation range0
+
       if warn then
          for (pn, _pdoc) in comment.Parameters do
             if not (tnames.Contains (Some pn)) then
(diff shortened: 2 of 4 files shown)


src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -78,6 +78,13 @@ module internal Utils =
     module Html =
         let sepWith s l = l |> List.sepWith (!! s) |> span []
 
+    type System.Xml.Linq.XElement with
+        member x.TryAttr (attr: string) =
```
baronfel, line 82: I added this for convenience and then only used it in the new code, but I'd of course be willing to reuse it in the existing attribute fetches in this file if you like.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -1560,8 +1567,12 @@ module internal SymbolReader =
             let exampleNodes = doc.Elements(XName.Get "example") |> Seq.toList
             [ for (id, e) in List.indexed exampleNodes do
                 let html = new StringBuilder()
-                let n = if id = 0 then "example" else "example-" + string id
-                rawData.[n] <- e.Value
+                let exampleId =
+                    match e.TryAttr "id" with
```
baronfel, line 1571: tied to the question in the [RFC change](https://github.com/fsharp/fslang-design/pull/630), what level of uniqueness should be enforced by tooling?

dsyme: Ideally fail if `--strict` is used, and warn otherwise

dsyme:
@baronfel Thanks! Could you add a test please?

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
baronfel: I'm noodling around the implementation a bit here still for validation purposes: if the scope is just within a particular example then that's easy to do validation on, because we have all the examples locally, but at any other scope than that things get hard because we don't have that data available and/or we'd have to push the validation up a level...

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
```diff
@@ -215,7 +215,7 @@ type HtmlRender(model: ApiDocModel) =
 
                   for e in m.Comment.Examples do
                       h5 [Class "fsdocs-example-header"] [!! "Example"]
-                      p [Class "fsdocs-example"] [embed e]
+                      p [Class "fsdocs-example"; if e.Id.IsSome then Id e.Id.Value ] [embed e]
```
baronfel, line 218: id's generated in the html now, yay

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -286,6 +295,30 @@ type ApiDocMember (displayName: string, attributes: ApiDocAttribute list, entity
               ParameterType=pty
               ParameterDocs=tdocs.TryFind pnm |} ]
 
+  do
+    let knownExampleIds =
+        comment.Examples
+        |> List.choose (fun x -> x.Id)
+        |> List.countBy id
+    for (id, count) in knownExampleIds do
```
baronfel, line 303: this pass checks for duplicate example ids on a per-member basis

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -286,6 +295,30 @@ type ApiDocMember (displayName: string, attributes: ApiDocAttribute list, entity
...
+        if count > 1 then
+            if warn
+            then
+                printfn "%s(%d,%d): warning: duplicate id for example '%s'" m.FileName m.StartLine m.StartColumn id
+            else
+                printfn "%s(%d,%d): error: duplicate id for example '%s'" m.FileName m.StartLine m.StartColumn id
+    for (id, _count) in knownExampleIds do
+        if id.StartsWith "example-" then
```
baronfel, line 311: this pass checks for missing example ids where we potentially inserted a dummy id

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
baronfel: settled on duplicate within a member at the moment, but we could expand that to the type semi-easily if desired

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
dsyme: `then` on same line as `if` please per style guide

dsyme (approved):

PR #711, opened 2021-10-09
Websocket CPU efficiency improvements.

amcguier:
I see high cpu usage using the `fsdocs watch` command in current
release version. It appears that the loop inside the socket is running
constantly since it never yields to receive socket messages from the
client, causing the process to use 100% of available CPU. Also, since
the reload creates a new socket connection (and suave creates a new
handler) this compounds since a new infinite loop is created after
every reload.

I've replaced the mutable boolean with a ManualResetEvent which all
connected sockets block on until a reload is called for. Once the
socket connection is closed the old handler exits. I chose
ManualResetEvent over AutoResetEvent because we want all connected
sockets to restart at the same time, rather than individually, since
someone might have multiple tabs open.

### DIFF
--- a/src/FSharp.Formatting.CommandTool/BuildCommand.fs
+++ b/src/FSharp.Formatting.CommandTool/BuildCommand.fs
@@ -290,21 +290,15 @@ module Serve =
         tag.Replace("{{PORT}}", string port)
 
 
-    let mutable signalHotReload = false
+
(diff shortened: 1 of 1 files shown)


dsyme:
@amcguier It looks good. Could you outline the manual testing you did?  Thanks

Issue #714, opened 2021-10-29
FSharp.Formatting package contains MSBuild assemblies


rainersigwald:
While helping investigate test failures in https://github.com/fsharp/FsAutoComplete/pull/825, I noticed that FSharp.Formatting contains several MSBuild assemblies:

* Microsoft.Build.Framework.dll
* Microsoft.Build.Tasks.Core.dll
* Microsoft.Build.Utilities.Core.dll
* Microsoft.NET.StringTools.dll

That caused some surprising behavior, since MSBuild assemblies need to be located from Visual Studio or a .NET SDK rather than redistributed with an application, because of their entanglement with targets and tasks that are part of the SDK. [MSBuildLocator docs](https://docs.microsoft.com/visualstudio/msbuild/updating-an-existing-application?view=vs-2019) have some details.

Can those files be elided from the package?

nhirschey:
Closed with https://github.com/fsprojects/FSharp.Formatting/pull/715

PR #715, opened 2021-10-30
try to remove MSBuild assemblies

nhirschey:
This is an attempt to remove MSBuild assemblies from the FSharp.Formatting.dll nuget. This change is requested in issue https://github.com/fsprojects/FSharp.Formatting/issues/714, which came up in @baronfel's attempt to upgrade Ionide/FSAC for dotnet 6. 

The changes are small, but I am a novice with respect to fsproj files and builds, so somebody with better knowledge should double check that I'm not doing anything stupid.

### Changes
1. I removed the explicit assembly inclusion in [FSharp.Formatting.fsproj](https://github.com/fsprojects/FSharp.Formatting/blob/master/src/FSharp.Formatting/FSharp.Formatting.fsproj#L14) that @baronfel  mentioned in the above referenced issue 714.

2. I set the MsBuild nugets in `paket.dependencies` to `copy_local: false` which "is equivalent to setting `ExcludeAssets=runtime` for a PackageReference in NuGet" ([relevant paket docs](http://fsprojects.github.io/Paket/nuget-dependencies.html#Controlling-whether-assemblies-should-be-copied-to-the-output-directory-during-build)). `ExcludeAssets=runtime` is the suggestion in the msbuild documentation link that @rainersigwald provided.  

### Testing:

1. I ran `dotnet fake build -t All` and all tests pass.

2. I inspected the internals of `artifacts\FSharp.Formatting.11.4.4.nupkg\lib\netstandard2.1\`. It included MsBuild dlls before the change, but after the change it does not have those dlls.

3. I inspected the command line tool nupkg and it still includes the msbuild dlls in `artifacts\FSharp.Formatting.CommandTool.11.4.4.nupkg\tools\net5.0\any\`

image

4. I tried using `artifacts/fsdocs.exe` to build a library of my own and to build the FSharp.Formatting docs itself. Everything worked fine. Below is a gif showing the results of building the libraries own documentation with `artifacts/fsdocs.exe watch --eval --properties Configuration=Release --clean`

no-msbuild

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -289,14 +289,16 @@
     <PropertyGroup>
       <PaketProjectFile>$(MSBuildProjectDirectory)/$(MSBuildProjectFile)</PaketProjectFile>
       <ContinuePackingAfterGeneratingNuspec>true</ContinuePackingAfterGeneratingNuspec>
+      <UseMSBuild16_10_Pack>false</UseMSBuild16_10_Pack>
+      <UseMSBuild16_10_Pack Condition=" '@(MSBuildMajorVersion)' >= '16' AND '@(MSBuildMinorVersion)' > '10' ">true</UseMSBuild16_10_Pack>
       <UseMSBuild16_0_Pack>false</UseMSBuild16_0_Pack>
-      <UseMSBuild16_0_Pack Condition=" '@(MSBuildMajorVersion)' >= '16' ">true</UseMSBuild16_0_Pack>
+      <UseMSBuild16_0_Pack Condition=" '@(MSBuildMajorVersion)' >= '16' AND (! $(UseMSBuild16_10_Pack)) ">true</UseMSBuild16_0_Pack>
       <UseMSBuild15_9_Pack>false</UseMSBuild15_9_Pack>
       <UseMSBuild15_9_Pack Condition=" '@(MSBuildMajorVersion)' == '15' AND '@(MSBuildMinorVersion)' > '8' ">true</UseMSBuild15_9_Pack>
       <UseMSBuild15_8_Pack>false</UseMSBuild15_8_Pack>
-      <UseMSBuild15_8_Pack Condition=" '$(NuGetToolVersion)' != '4.0.0' AND (! $(UseMSBuild15_9_Pack)) AND (! $(UseMSBuild16_0_Pack)) ">true</UseMSBuild15_8_Pack>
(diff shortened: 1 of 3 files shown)


dsyme:
This looks great!!!

baronfel:
Perfect, that's exactly the changes required. Nice job doing validation as well 👍

dsyme:
I'm not sure this is a sound fix - I'm getting failures reporting missing MSBuild DLLs when running tests locally.

FCS has had an MSBuild dependency for a long time.  What do we know about why that's currently causing a problem for FsAutoComplete?

baronfel:
the dependency isn't the problem, it's that the dlls themselves were xcopied into the library. the dependencies should be expressed as nuget package dependencies (which is handled by the FSharp.Compiler.Service dependency itself), and not be inlined into the package output itself.  Applications that need the dlls are probably running into a PackageReference transitive dependency copy issue and might be resolved by adding an explicit top-level dependency to their paket.references on the MSBuild nugets required. I think there's and SDK or Nuget issue about this, and it's bitten me a few times as well.

dsyme:
I see.

Yes, they shouldn't be copied into the package.  I was meaning the `copy_local: false` part which is about inside this repo. Thanks

nhirschey:
1. At the time I made this pull request, I ran `dotnet fake build -t All` locally and all tests passed. However, today I am seeing test failures when I run that too. The only difference I can think of is that in that time I have installed Visual Studio 2022 on my machines this past week.

2. When I use `global.json` to pin to dotnet 5.0, I can still pass tests:

```json
{
    "sdk": {
      "version": "5.0.200",
      "rollForward": "minor"
    }
  }
```

image


3. The `copy_local:false` was added to follow this guide: https://docs.microsoft.com/en-us/visualstudio/msbuild/updating-an-existing-application?view=vs-2019#use-nuget-packages-preferred

4. I believe the project cracker in this project uses Ionide's project cracker. On dotnet 6, could it be hitting the same msbuild issues that @baronfel  is working through in ionide?

baronfel:
It might be worth trying to see if version 0.54.2 solves this for you, then. We changed how we detect and isolate the msbuild versions in use.  0.55.0 also bumps to FCS 41, if you want to go that far.

dsyme:
> It might be worth trying to see if version 0.54.2 solves this for you, then. We changed how we detect and isolate the msbuild versions in use. 0.55.0 also bumps to FCS 41, if you want to go that far.

In #719 I'm updating to .NET 6, but it's slow going, the combined bump to ProjInfo 0.55 and FCS 41 and .NET 6 is tricky https://github.com/ionide/proj-info/issues/122

One issue was FAKE didn't have net6.0 support, I think that's being fixed.

PR #750, opened 2022-06-02
Add support for additional mime-types.

### LINKED ISSUE #749: Video content is not being served.
Hello, 

I'm trying to include a video on a static HTML page.
The videos (*.mp4 and *.webm) are currently not playing and I believe that is because the mime types are not set by Suave.

https://github.com/fsprojects/FSharp.Formatting/blob/f389c0df2355412286b7a05a4aa86992f878b46d/src/fsdocs-tool/BuildCommand.fs#L630-L649

https://github.com/SuaveIO/suave/blob/57d69e39e67a5b1a1c60b7c015c698a1901a9420/src/Suave/Combinators.fs#L90-L114

The Suave docs explain how to add an additional mime type: https://suave.io/files.html

Can I include some more mime types? Hardcoded or via settings?

nojaf:
Fixes #749.

### DIFF
--- a/docs/commandline.md
+++ b/docs/commandline.md
@@ -72,12 +72,13 @@ report an error (e.g. "Problem loading...", "Connection was reset").
 
  Restarting may be necesssary on changes to project files. The same parameters are accepted, plus these:
 
-| Command Line Option                 |  Description    |
-|:-----------------------|:-----------------------------------------|
-| `--noserver`     |   Do not serve content when watching.  |
-| `--nolaunch`     |   Do not launch a browser window. |
-| `--open`     |   URL extension to launch http://localhost:<port>/%s. |
-| `--port`     |   Port to serve content for http://localhost serving. |
+| Command Line Option                 | Description                                                     |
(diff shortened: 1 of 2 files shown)


baronfel:
this is a good escape hatch, but should the default set of MIME times be expanded? `.mp4` is pretty common these days

nojaf:
Perhaps, but that begs the question should it be done here or in Suave itself?

baronfel:
I think it could be a two-pronged approach - fix here as an intermediate, and provide the same fix upstream to Suave to let them decide if they want to take it.

nojaf:
Sounds reasonable, I would need the escape hatch for the more obscure stuff in my project.

nojaf:
@dsyme
, @eiriktsarpalis how do you want to see this fixed?

dsyme:
> how do you want to see this fixed?

I think expanding the list here to include all the entries from ASP.NET Core makes sense, and then also contributing that to Suave.

nojaf:
Closing in favour of #752.

PR #752, opened 2022-06-08
Add same mime-types as ASP.NET Core.

nojaf:
Alternative to #750.
@dsyme this solves my original problem as well, so we might not yet need the additional cli option.
I also opened https://github.com/SuaveIO/suave/issues/764.


dsyme:
👍

Issue #754, opened 2022-07-14
Customize the menu items of {{fsdocs-list-of-documents}} and {{fsdocs-list-of-namespaces}}


nojaf:
Hello, we are currently experimenting with FsDoc. We aim to have our own template and are experimenting with Bootstrap 5, Sass and some other ideas.

We are currently hitting a limitation that:
```fsharp
{{fsdocs-list-of-documents}}
{{fsdocs-list-of-namespaces}}
```

is generating some HTML that doesn't fit our needs.
Would there be a way to have some control over what this outputs?

//cc @yisusalanpng

dsyme:
Certainly, you can suggest something?

nojaf:
@dsyme, I made an experiment with transforming the menu information in an outside endpoint.
https://github.com/fsprojects/FSharp.Formatting/commit/a4d9f8f3fa3acbdb2337a57e35ebab4b48cccbac

I made a small script to act as the receiving endpoint:
```fsharp
#r "nuget: Suave"
#r "nuget: Thoth.Json.Net, 8.0.0"
#r "nuget: Fable.React, 8.0.1"

open System.Net
open Suave
open Suave.Filters
open Suave.Operators
open Suave.Successful
open Thoth.Json.Net

type Item = {
    Title: string
    Category: string
    CategoryIndex: int
    Index: int
    Link: string
}

let decodeStringAsInt = Decode.string |> Decode.map (int)

let decodeItem: Decoder<Item> =
    Decode.object (fun get -> {
        Category = get.Required.Field "category" Decode.string
        CategoryIndex = get.Required.Field "categoryIndex" decodeStringAsInt
        Index = get.Required.Field "index" decodeStringAsInt
        Link = get.Required.Field "link" Decode.string
        Title = get.Required.Field "title" Decode.string
    }
    )

open Fable.React
open Fable.React.Props

let view (items: Item array) : string =
    let groups = Array.groupBy (fun i -> i.CategoryIndex) items

    let children =
        groups
        |> Array.map (fun (_, groupItems) ->
            let groupTitle = groupItems[0].Category

            let id = $"menu-{groupTitle}-collapse".Replace(" ", "-").Trim().ToLower()

            let groupItems =
                groupItems
                |> Array.map (fun (gi: Item) ->
                    li [] [ a [ Href gi.Link; ClassName "ms-4 my-2 d-block" ] [ str gi.Title ] ]
                )

            li [ ClassName "mb-1" ] [
[...]

PR #758, opened 2022-08-03
Proposal for custom templating

alanlomeli:
Related to #754 
FYI: @nojaf


src/FSharp.Formatting.Common/Menu.fs
```diff
@@ -0,0 +1,37 @@
...
+open System.IO
+open FSharp.Formatting.Templating
+
+type MenuItem =
+    {
+        Link: string
+        Content: string
+    }
+//am I suppose to have this global lets?
+let fsdocsMenuHeaderContentKey = ParamKey "fsdocs-menu-header-content"
```
nojaf, line 11: You probably want to move these to https://github.com/fsprojects/FSharp.Formatting/blob/139754a685d26badb19c38998f88cc8e1de7dc87/src/FSharp.Formatting.Common/Templating.fs#L29

src/FSharp.Formatting.Common/Menu.fs
```diff
@@ -0,0 +1,37 @@
...
+let fsdocsMenuItemsKey = ParamKey "fsdocs-menu-items"
+let fsdocsMenuItemLinkKey = ParamKey "fsdocs-menu-item-link"
+let fsdocsMenuItemContentKey = ParamKey "fsdocs-menu-item-content"
+let pwd = Directory.GetCurrentDirectory()
```
nojaf, line 15: I would either make `pwd` private or move it into both the public functions.

src/FSharp.Formatting.ApiDocs/GenerateMarkdown.fs
```diff
@@ -362,9 +362,19 @@ type MarkdownRender(model: ApiDocModel) =
...
-        listOfNamespacesAux otherDocs nav nsOpt
-        |> List.map (fun html -> html.ToString())
-        |> String.concat "             \n"
+        let isTemplatingAvailable =
+            match menuTemplateFolder with
+            | None -> false
+            | Some input -> FSharp.Formatting.Common.Menu.isTemplatingAvailable input
+
+        if isTemplatingAvailable then
+            listOfNamespacesAux otherDocs nav nsOpt
```
nojaf, line 371: You should call `createMenu` here instead.

docs/_menu_template.html
```diff
@@ -0,0 +1,4 @@
+<li class="nav-header">
```
nojaf, line 1: I also don't think you should add these templates for the docs of this repository.
We are proposing an alternative for the default, so having a template that mimics the default is a bit strange.

FSharp.Formatting.sln
```diff
@@ -36,6 +36,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{312E452A-1
 		docs\styling.md = docs\styling.md
 		docs\upgrade.md = docs\upgrade.md
 		docs\_menu-item_template.html = docs\_menu-item_template.html
+		docs\_menu_template.html = docs\_menu_template.html
```
nojaf, line 39: These should probably be removed as well.

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
```diff
@@ -618,10 +618,27 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
         let isTemplatingAvailable =
             match menuTemplateFolder with
             | None -> false
-            | Some input -> FSharp.Formatting.Common.Menu.isTemplatingAvailable input
+            | Some input -> Menu.isTemplatingAvailable input
 
         if isTemplatingAvailable then
-            "TODO!"
+                 if otherDocs && model.Collection.CollectionName <> "FSharp.Core" then
```
nojaf, line 624: Please format your code before committing 😉.

yisusalanpng: Sure ! Any formatter you recommend?

nojaf: I'd use the one they check in the CI, https://github.com/fsprojects/FSharp.Formatting/blob/main/.github/workflows/pull-requests.yml#L26

src/FSharp.Formatting.Common/Templating.fs
```diff
@@ -114,6 +114,10 @@ module ParamKeys =
 
     /// A parameter key known to FSharp.Formatting
     let ``fsdocs-watch-script`` = ParamKey "fsdocs-watch-script"
+    let ``fsdocsMenuHeaderContentKey`` = ParamKey "fsdocs-menu-header-content"
```
nojaf, line 117: Please use the exact same format as the existing code.
This is a general guideline for any PR you ever make, look at the existing code and just do the same thing.

```
/// A parameter key known to FSharp.Formatting
let ``fsdocs-menu-header-content`` = ParamKey "fsdocs-menu-header-content"
```

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
```diff
@@ -618,10 +618,27 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
...
+                    let menuItems =
+                     let title = "All Namespaces"
+                     let link = model.IndexFileUrl(root, collectionName, qualify, model.FileExtensions.InUrl)
+                     [{Menu.MenuItem.Link = link
+                       Menu.MenuItem.Content = title }]
+                    Menu.createMenu menuTemplateFolder.Value "API Reference" menuItems
+                 else
+                    let headerName =
+                        let categorise = Categorise.model model
```
nojaf, line 633: I think you more or less want something like this:
```fsharp
                    let categorise = Categorise.model model
                    if categorise.Length = 0 then
                        ""
                    else
                        let menuItems =
                            categorise
                            |> List.map (fun (_, ns) ->
                                 { Link = ns.Url(root, collectionName, qualify, model.FileExtensions.InUrl)
                                   Content = ns.Name }
                            )

                        createMenu menuTemplateFolder.Value "Namespaces" menuItems
```

src/FSharp.Formatting.Common/FSharp.Formatting.Common.fsproj
```diff
@@ -14,6 +14,12 @@
     <Compile Include="PynbModel.fs" />
     <Compile Include="HtmlModel.fs" />
     <Compile Include="Templating.fs" />
+    <Compile Include="Menu.fs" />
+  </ItemGroup>
+  <ItemGroup>
```
nojaf, line 19: Any idea if this is necessary? Seems a bit odd.

docs/styling.md
```diff
@@ -3,15 +3,16 @@ category: Documentation
 categoryindex: 1
 index: 7
 ---
-# Customization and Styling 
+
```
nojaf, line 6: Please revert all changes to this file that are not new content. If you want to clean up something you can do this in another PR. The goal of this one is to add a custom menu structure, any other change is just confusing noise. The git history will be confusing because of this.
You always want to make a PR as small as possible. Provide minimum changes to solve the problem and don't touch anything else.

docs/styling.md
```diff
@@ -71,63 +72,81 @@ As an example, here is [a page with `fsdocs-navbar-position` set to `fixed-right
...
 with the existing default template.
 
 > NOTE: To enable hot reload during development with `fsdocs watch` in a custom `_template.html` file,
-> make sure to add the single line `{{fsdocs-watch-script}}`  to your `<head>` tag.
+> make sure to add the single line `{{fsdocs-watch-script}}` to your `<head>` tag.
 
 > NOTE: There is no guarantee that your template will continue to work with future versions of F# Formatting.
 > If you do develop a good template please consider contributing it back to F# Formatting.
 
+## Customizing menu items by template
```
nojaf, line 122: This really is the only change you want to have in this file.

docs/styling.md
```diff
@@ -71,63 +72,81 @@ As an example, here is [a page with `fsdocs-navbar-position` set to `fixed-right
...
+
+You can add advanced stylings to the sidebar generated menu items by creating a new template for it.
+To customize the generated headers, add template file `docs/_menu_template.html`. A starting template with the default configuration would look like:
```
nojaf, line 125: Mention they should put a file in the `--input` folder, by default that is typically the `docs` folder but it could be something different.

Both templates should exist in order for this to work, please add a note for this as well.

docs/_menu-item_template.html
```diff
@@ -0,0 +1 @@
+<li class="nav-item"><a href="{{fsdocs-menu-item-link}}" class="nav-link">{{fsdocs-menu-item-content}}</a></li>
```
nojaf, line 1: Don't add these files to the `docs` folder, again the website https://fsprojects.github.io/FSharp.Formatting does not use this. It only adds confusing that these files are here.

src/FSharp.Formatting.Common/Menu.fs
```diff
@@ -0,0 +1,56 @@
...
+open System
+open System.IO
+open FSharp.Formatting.Templating
+
+type MenuItem = { Link: string; Content: string }
+
+let private snakeCase (v: string) =
+    let replace = v.Replace(" ", "")
+
+    System
```
nojaf, line 12: I didn't catch this at first but spaces are not replaced by this code.
Leads to identifiers with spaces if the input has any.
Add `.Replace(" ", "")` before `.ToLower()`.

docs/styling.md
yisusalanpng: Sorry Nojaf, this one got a little out of hand. I don't know why my IDE made all of those changes.

docs/styling.md
```diff
@@ -118,6 +118,27 @@ with the existing default template.
 > NOTE: There is no guarantee that your template will continue to work with future versions of F# Formatting.
 > If you do develop a good template please consider contributing it back to F# Formatting.
 
+## Customizing menu items by template
+
+You can add advanced stylings to the sidebar generated menu items by creating a new template for it.
+For adding a template via command line you can use the `--input` option or alternatively add them to the docs/ folder.
```
nojaf, line 124: This is slightly confusing. What you basically want to say is that the `fsdocs` will look for menu templates in the `--input` folders. This folder defaults to `docs`.
You are not really adding any template via the command line.

Also, make it clear that both templates need to exist before. Otherwise, it won't work.

docs/styling.md
```diff
@@ -118,6 +118,27 @@ with the existing default template.
...
+
+To customize the generated headers, use file `_menu_template.html` with starting template:
+
+```
```
nojaf, line 128: Use <pre>```html</pre> here.

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -550,40 +556,64 @@ type internal DocContent
...
-                  match cat with
-                  | Some c -> li [ Class "nav-header" ] [ !!c ]
-                  | None -> li [ Class "nav-header" ] [ !! "Other" ]
-
-                  for model in modelsInCategory do
+        if FSharp.Formatting.Common.Menu.isTemplatingAvailable input then
+            let createGroup (header: string) (items: LiterateDocModel list) : string =
+                //convert items into menuitem list
+                let menuItems =
+                    items
```
nojaf, line 563: The sort order of the menu items is not yet respected.

image

This is alphabetical now, the original code sorts it like this:
```fsharp
|> List.sortBy (fun model ->
                              match model.Index with
                              | Some s ->
                                  (try
                                      int32 s
                                   with _ ->
                                       Int32.MaxValue)
                              | None -> Int32.MaxValue)
```

Extract the existing sort to a local helper function and re-use it in the new code.

src/FSharp.Formatting.Common/Menu.fs
```diff
@@ -0,0 +1,55 @@
+module FSharp.Formatting.Common.Menu
+
+open System
+open System.IO
+open FSharp.Formatting.Templating
+
+type MenuItem = { Link: string; Content: string }
+
+let private snakeCase (v: string) =
```
nojaf, line 9: This still produces kinda funky identifiers.
Consider "End-user documentation" being transformed to "_end-userdocumentation".
I would trim the leading and trailing `_` and replace spaces with `-` or `_` instead.
And may even replace all `_` to `-` or vice versa, so you don't have a mix.

Feel free to revisit this entire helper function if that makes more sense.

nojaf (approved):
Looks good, @dsyme this is ready for review.

dsyme:
Great!

PR #760, opened 2022-08-08
Preserve underscores in emphasis.

### LINKED ISSUE #389: Markdown parser gets multiple-underscores-inside-italics wrong
The markdown parser gets underscores-inside-italics wrong. For instance,
```
A *asterisk_one_two* B
```

On both GitHub and CommonMark, it's rendered as an italic with underscores:

A *asterisk_one_two* B

But FSharp.Formatting.Markdown parses it as a nested emphasis:
```
Emphasis [ Literal "asterisk", Emphasis [ Literal "one"], Literal "two"]
```

(Actually, to be honest, I don't know what is the correct behavior. I'm assuming that CommonMark is correct here.)

Here are some more examples:
```
A *asterisk_one_two* B

A *asterisk_one_two_three* B

A *asterisk_one_two_three_four* B

A _underscore_one_two_ B

A _underscore_one_two_three_ B

A _underscore_one_two_three_four_ B
```

A *asterisk_one_two* B

A *asterisk_one_two_three* B

A *asterisk_one_two_three_four* B

A _underscore_one_two_ B

A _underscore_one_two_three_ B

A _underscore_one_two_three_four_ B

nojaf:
Fixes #389.
This isn't an airtight fix but will work in the most common situations.
It worked for the Fantomas documentation.

### DIFF
--- a/src/FSharp.Formatting.Markdown/MarkdownParser.fs
+++ b/src/FSharp.Formatting.Markdown/MarkdownParser.fs
@@ -79,7 +79,21 @@ let (|DelimitedMarkdown|_|) bracket input =
     let rec loop acc =
         function
         | EscapedChar (x, xs) -> loop (x :: '\\' :: acc) xs
-        | input when List.startsWith endl input -> Some(List.rev acc, input)
+        | input when List.startsWith endl input ->
+            let rest = List.skip bracket.Length input
+
+            match rest with
+            | []
+            | (' '
+            | '\r'
+            | '\n'
+            | '.'
+            | '?'
+            | '!') :: _ -> Some(List.rev acc, input)
+            | rest ->
+                let head = List.take bracket.Length input
+                loop [ yield! head; yield! acc ] rest
+
         | x :: xs -> loop (x :: acc) xs
         | [] -> None
     // If it starts with 'startl', let's search for 'endl'
@@ -89,7 +103,6 @@ let (|DelimitedMarkdown|_|) bracket input =
         | None -> None
     else
         None
-
 
 /// This is similar to `List.Delimited`, but it skips over Latex inline math characters.
 let (|DelimitedLatexDisplayMath|_|) bracket input =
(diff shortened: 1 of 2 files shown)


nhirschey:
As written you will fail markdown emphasis near commas, semi-colons, etc. such as in the below test (you can test the spec [here](https://spec.commonmark.org/dingus/)). I suggest a fix below, a little more tight but still not perfect.

```fsharp
[<Test>]
let ``Underscore inside italic and bold near punctuation is preserved`` () =
    let doc = "This is **bold_bold**, and this _italic_; and _this_too_: again."

    let expected =
        "<p>This is <strong>bold_bold</strong>, and this <em>italic</em>; and <em>this_too</em>: again.</p>\r\n"
        |> properNewLines

    Markdown.ToHtml doc |> shouldEqual expected

(*
Actual:
"<p>This is **bold_bold**, and this _italic_; and _this_too_: again.</p>"
*)
```

It will pass if you instead don't end emphasis with underscores that come before letters or numbers:

```fsharp
/// Succeeds when the specificed character list starts with a letter or number
let inline (|AlphaNum|_|) input =
    let re = """^[a-zA-Z0-9]"""
    let match' = Regex.Match(Array.ofList input |> String, re)
    if match'.Success then
        let entity = match'.Value
        let _, rest = List.splitAt entity.Length input
        Some(char entity, rest)
    else
        None

/// Matches a list if it starts with a sub-list that is delimited
/// using the specified delimiters. Returns a wrapped list and the rest.
///
/// This is similar to `List.Delimited`, but it skips over escaped characters.
let (|DelimitedMarkdown|_|) bracket input =
    let _startl, endl = bracket, bracket
    // Like List.partitionUntilEquals, but skip over escaped characters
    let rec loop acc =
        function
        | EscapedChar (x, xs) -> loop (x :: '\\' :: acc) xs
        | input when List.startsWith endl input ->
            let rest = List.skip bracket.Length input
            match rest with
            | AlphaNum (x, xs ) -> loop (x :: endl @ acc) xs
            | _ -> Some (List.rev acc, input)
        | x :: xs -> loop (x :: acc) xs
        | [] -> None
    // If it starts with 'startl', let's search for 'endl'
    if List.startsWith bracket input then
        match loop [] (List.skip bracket.Length input) with
        | Some (pre, post) -> Some(pre, List.skip bracket.Length post)
        | None -> None
    else
        None
```

nojaf:
Thanks!

dsyme (approved):

PR #763, opened 2022-08-19
Pass more commonmark emphasis parsing tests

nhirschey:
This PR improves parsing of markdown emphasis with * and _.

It adds 13 additional emphasis parsing tests from the commonmark [spec](https://github.com/commonmark/commonmark-spec/blob/48ba84093ec8ef1f8578517b5d14efa4f932124b/spec.txt#L6099). 8 of them currently fail on the main branch (first push will show that).

The new code in MarkdownParser.fs passes all old and new tests.

### DIFF
--- a/src/FSharp.Formatting.Markdown/MarkdownParser.fs
+++ b/src/FSharp.Formatting.Markdown/MarkdownParser.fs
@@ -69,17 +69,135 @@ let inline (|EscapedLatexInlineMathChar|_|) input =
     | '\\' :: (('$') as c) :: rest -> Some(c, rest)
     | _ -> None
 
-/// Succeeds when the specified character list starts with a letter or number
-let inline (|AlphaNum|_|) input =
-    let re = """^[a-zA-Z0-9]"""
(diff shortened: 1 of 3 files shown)


src/FSharp.Formatting.Markdown/MarkdownParser.fs
```diff
@@ -69,17 +69,127 @@ let inline (|EscapedLatexInlineMathChar|_|) input =
...
+    match input with
+    | ('*'
+    | '_') :: _tail as (h :: t) ->
+        let run, rest = List.partitionWhile (fun x -> x = h) (h :: t)
+        Some(run, rest)
+    | _ -> None
+
+/// Succeeds when there's a match to a string of * or _ that could
+/// open emphasis.
+let inline (|LeftDelimiterRun|_|) input =
```
dsyme, line 121: I'll remove the inline - these are too big to inline and there will be no useful performance gain from using it

dsyme (approved):

dsyme:
Great tests, thanks!

PR #762, opened 2022-08-15
Add HTML code element for code samples to enable syntax highlight.

yazeedobaid:
## Motivation
Enabling syntax highlights on XML documentation code samples will make it easy for readers and users of an API to understand the code samples quickly and without much effort. The current published version of `fsdocs` enable syntax highlight on comments written in markdown format. However, following the guidelines in [F# Recommended XML doc extensions for F# documentation tooling RFC](https://github.com/fsharp/fslang-design/blob/main/tooling/FST-1031-xmldoc-extensions.md) produces an output that has no syntax highlight nor we can add syntax highlight to it due to the HTML structure that the samples are written on, the structure is not the convention so, other libraries will not detect it.

Following the conventions from syntax highlighting libraries like [highlightjs](https://highlightjs.org/) and [prismjs](https://prismjs.com/) and [w3.org](https://www.w3.org/) site samples, the code samples should be put in a `code` element inside the `pre` element.

This PR attempt to add this to enable syntax lighlight.

Fixes #761

## Example

For example, for the following standard XML documentation code:

```F#
/// <summary>
/// A test module...
/// <example>
/// <code lang="fsharp">
///         let files =
///                [ "build"; "docs" ]
///                |> List.map (fun n -> sprintf "%s.zip" n)
/// </code>
/// </example>
/// </summary>
```

The result after using the tool to generate API documentation is as follows:

```diff
<pre>
+    <code class="fsharp language-fsharp">
        let files =
            [ "build"; "docs" ]
            |> List.map (fun n -> sprintf "%s.zip" n)
+    </code>
</pre>
```

Note that the language on XML `code` element has transferred and been mapped to the corresponding `HTML` element for syntax highlighters to get and operate on. In this way, users can pull a syntax highlighter and it will work out of the box.

By this, the XML documentation will be kept with no changes. However, the output HTML will be as follows with changes highlighted.

PR #780, opened 2022-11-18
Use fssnip highlighter in API docs code examples.

nhirschey:
This adds syntax highlighting to API docs code examples. PR #762 from @yazeedobaid was solving the same problem, but it relied on external tools such as highlight.js to do the highlighting and required users to add that dependency. This PR uses FSharp.Formatting's baked-in highlighter to achieve the same goal.

For example, building the F# core docs [Array.averageBy](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-arraymodule.html#averageBy) example using this PR:

image

----------------------

In comparison, this is how it looks now:

image

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+## 17.1.0
+* [Add syntax highlighting to API docs](https://github.com/fsprojects/FSharp.Formatting/pull/780)
+
 ## 17.0.0
 
 * Update to .NET 7.0.100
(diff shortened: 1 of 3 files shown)


nhirschey:
@yazeedobaid are you happy with this version?

nhirschey:
Now with the extra trailing blank line after the last line of code removed (adding back a prior dsyme fix). Should be ready for release.

image

yazeedobaid:
@nhirschey tested the branch with the FAKE website and all is good. But I may add back that padding by a CSS rule for the code examples since it is more visually  looking good with some padding:
image
 
That is FAKE related. Thanks for the update.

PR #783, opened 2022-12-05
fix apidocs markdown comment

nhirschey:
Typo, was "(*" instead of "(**". As a result the *Go to Source links* and *Markdown Comments* section of the api docs was rendering as a regular code comment.

This typo was created in PR #600. Prior to that PR it started with "(**".

This only affects the docs, no new release of the tool is needed.


nojaf:
Thanks!

Issue #786, opened 2023-01-08
[omit] vs <exclude/>


davedawkins:
```sh
% dotnet fsdocs --version
fsdocs 16.1.1
% dotnet --version
6.0.404
```
Both `[omit]` and `<exclude>` are mentioned in the [documentation](https://fsprojects.github.io/FSharp.Formatting/apidocs.html).

However:
- `<exclude>` doesn't appear to work (neither does `<exclude/>`)
- `[omit]` *does* work, but:
  - it produces a warning
  - sometimes it doesn't and becomes the summary for the item being excluded. 

The warning (which gives a good example of how to use `<exclude/>`:
```sh
The use of `[omit]` and other commands in XML comments is deprecated, please use XML extensions, see https://github.com/fsharp/fslang-design/blob/master/tooling/FST-1031-xmldoc-extensions.md
```

When `[omit]` doesn't work:
<img width="326" alt="image" src="https://user-images.githubusercontent.com/285421/211220460-4a0c908e-2b98-4c29-b845-3e0f219abe02.png">

davedawkins:
For reference, BindHelpers is an internal module. Eg
```fs
[<AutoOpen>]
module Sutil.Bindings

///[omit]
[<AutoOpen>]
module BindHelpers =
    // ..
```

davedawkins:
Verified that it doesn't work in 17.2 either
```
% dotnet --version
7.0.101
% dotnet fsdocs --version
fsdocs 17.2.0
```

I have tried:
```
///<exclude>
/// <exclude>
///<exclude/>
/// <exclude/>
/// <exclude> </exclude>
/// <exclude value="true">true</exclude>
```

From your link:
```fs
        for e in doc.Elements(XName.Get "exclude") do
            cmds.["exclude"] <- e.Value

        for e in doc.Elements(XName.Get "omit") do
            cmds.["omit"] <- e.Value
```

My guess is that either:
- `exclude` isn't being parsed into `doc.Elements` or
- `e.Value` isn't set correctly for `<exclude/>` or
- `cmds ["exclude"]` is being ignored somewhere

If you're a maintainer, could you point me at the code that assigns/parses/builds `doc.Elements` please?

davedawkins:
I'm writing a test for `<exclude/>` and `[omit]`

nojaf:
I can't really say I know anything more than you do at this point.
[...]

Issue #785, opened 2023-01-08
Setting root for links generated by fsdocs


davedawkins:
This is how I'm doing it now, to test locally:
```xml
    <PackageProjectUrl>..</PackageProjectUrl>
    <!-- <PackageProjectUrl>https://github.com/DaveDawkins/Sutil</PackageProjectUrl> -->
```

in order to have `root` set as shown here
```
Substitutions/parameters:
  root --> ../
  fsdocs-authors --> David Dawkins
  fsdocs-collection-name --> Sutil
  fsdocs-collection-name-link --> ../
```

Is there  a better way? (reading the source suggests that it will derive `root` from `<PackageProjectUrl>` only

Thank you!

nhirschey:
You can specify parameters at the [command line](https://fsprojects.github.io/FSharp.Formatting/commandline.html).

```
dotnet fsdocs build --parameters root ../
```

```
Substitutions/parameters:
  root --> ../
```

davedawkins:
That is useful to know, thank you. Just to make sure I'm not msising anything, does it mention `root` in that page there, or how to rebase links? If not, then I can submit a doc PR

davedawkins:
Tested, that works - thanks again

nhirschey:
It's there but not super explicit and there's no examples. It could probably be improved. At the bottom of the [templates and substutions](https://fsprojects.github.io/FSharp.Formatting/content.html#Templates-and-Substitutions) section that you were referencing it mentions

> For the fsdocs tool, additional substitutions can be specified using --parameters.

The command line tool --help output does a better job documenting it, showing the "--parameters key value" format.

PR #787, opened 2023-01-10
Fixes for <exclude/> support, with tests and docs.

davedawkins:
Documentation on rebasing links
Issues #786 , #785


FSharp.Formatting.sln
```diff
@@ -83,6 +83,8 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "TestLib1", "tests\FSharp.Ap
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "TestLib2", "tests\FSharp.ApiDocs.Tests\files\TestLib2\TestLib2.fsproj", "{48EFFECF-ECB0-4DF3-A704-B56AB07557BF}"
 EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "TestLib3", "tests\FSharp.ApiDocs.Tests\files\TestLib3\TestLib3.fsproj", "{52B949AA-A3F7-4894-B713-804BAEB71118}"
```
nojaf, line 86: `TestLib3` isn't showing up under the right folder for me:
image

docs/apidocs.fsx
```diff
@@ -169,6 +169,63 @@ type GenericClass2<'T>() =
...
+
+If you want to exclude modules or functions from the API docs you can use the `<exclude/>` tag.
+It needs to be set on a separate triple-slashed line, and can either appear on its own or as part
+of an existing `<summary>` (for example, you may wish to hide existing documentation while it's in progress).
+The `<exclude/>` tag can be the first or last line in these cases.
+
+Some examples:
+*)
+
+/// </exclude>
```
nojaf, line 183: Should this not be `<exclude />`? `</exclude>` without opening tag doesn't strike me as valid XML.

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
```diff
@@ -707,7 +707,7 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
             printfn "  Generating %s" outFile
             SimpleTemplating.UseFileAsSimpleTemplate(substitutions, templateOpt, outFile)
 
-        for info in model.EntityInfos do
+        for info in model.EntityInfos |> List.filter (fun info -> not info.Entity.Exclude) do
```
nojaf, line 710: Should we consider having the filtering inside `.EntityInfos` instead?
Based on the usage of this member I think this makes sense.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -2261,12 +2261,30 @@ module internal SymbolReader =
...
+            // 1.
+            //  // Not triple-slash ccomment
+            //  /// <exclude/>
+            //
+            // 2.
+            //  /// <exclude/>
+            //
+            // So, we need to let the 'null' case handle this to extract the <exclude> if it's there
+            //
+            // | null when String.IsNullOrEmpty el.Value ->
```
nojaf, line 2277: These last two lines of the comment can be omitted?

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -2261,12 +2261,30 @@ module internal SymbolReader =
...
+            //     cmds, ApiDocComment.Empty, None
+
             | null ->
                 // We let through XML comments without a summary tag. It's not clear
                 // why as all XML coming through here should be from F# .XML files
                 // and should have the tag.  It may be legacy of previously processing un-processed
                 // XML in raw F# source.
+                //
+                // 9-Jan-23: See comment above for at least one reason why we pass through here now
```
nojaf, line 2286: This journal entry seems a tad excessive. You explained the reasoning not ten lines ago.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs-debug
```diff
@@ -0,0 +1,3157 @@
+namespace rec FSharp.Formatting.ApiDocs
```
nojaf, line 1: Should this file exist?

nojaf (changes requested):
This looks like a solid PR. I have some minor nitpicks along the way but I think we can take this in.
At the same time, I'm no expert so @nhirschey do you want to glance over this one as well?

@davedawkins if you update the `RELEASE_NOTES.md` file with your fix and a new version (say `17.2.1`) a new version would be released once we merged this to main.
If you don't need a new version that's fine as well.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs-debug
davedawkins: No - will fix

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
davedawkins: Well, it's a single line pointing at the explanation, in response to the question raised in the previous developer's comment "why are we here". How do you want to improve this?

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
davedawkins: I will take a look at EntifyInfos. My preference was to do it in the logic that calls `checkAccess`, but I believe that's before we've processed the comments.

docs/apidocs.fsx
davedawkins: Yes, a typo. Sorry :-/

FSharp.Formatting.sln
davedawkins: I will fix this.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
davedawkins: I removed all the comments. I see your point (to an extent), however: the original comment either needs a response (in context), or just removing. My approach is not to remove comments, but respond to them in a timeline to help out developers down the line.  I think however nothing is lost - if you think it's excessive, others may do too.

davedawkins:
All updates made as requested.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -2261,12 +2261,21 @@ module internal SymbolReader =
             let sum = el.Element(XName.Get "summary")
 
             match sum with
```
nojaf, line 2263: The match is looking a bit funny now.
Perhaps go with:
```fsharp
            // sum can be null with null/empty el.Value when an non-"<summary>" XML element appears
            // as the only '///' documentation command:
            //
            // 1.
            //  // Not triple-slash ccomment
            //  /// <exclude/>
            //
            // 2.
            //  /// <exclude/>
            if isNull sum then
                let doc, nsels = readXmlCommentAsHtmlAux false ctx.UrlMap el cmds

                let nsdocs = readNamespaceDocs ctx.UrlMap nsels
                cmds, doc, nsdocs
            elif ctx.MarkdownComments then
                readMarkdownCommentAndCommands ctx sum.Value el cmds
            else
                if sum.Value.Contains("<exclude") then
                    cmds.["exclude"] <- ""

                    printfn
                        "Warning: detected \"<exclude/>\" in text of \"<summary>\" for \"%s\". Please see https://fsprojects.github.io/FSharp.Formatting/apidocs.html#Classic-XML-Doc-Comments"
                        xmlSig

                readXmlCommentAndCommands ctx sum.Value el cmds
```
instead.

nojaf (approved):
Looks good to me.

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
davedawkins: Agreed, your fix applied

nhirschey:
Thanks so much @davedawkins. @nojaf sure, I should be able take a look in a few hours.

davedawkins:
Please hold off from applying this PR. I think I've found an issue with excluding `type`.

davedawkins:
There's no issue with excluding type as such, but the module will still list an excluded type. Additionally, the search index is listing excluded items, so I'm going to fix both those issues today. Unless requested otherwise, I'm going to check those changes into this branch (and PR).

nojaf:
Hello @davedawkins, thank you so much for your persistence here!

I think having the index changes in this PR is ok. I took a peek at the latest changes and everything is still ok for me.

davedawkins:
@nojaf, no problem. I'm determined to get Sutil's API documented! Thank you for help too.

nhirschey (approved):

nhirschey:
Wonderful set of fixes @davedawkins. Thank you! It looks good to me as well. 

I went ahead and merged the PR to trigger the release. It should hit the feed soon.

PR #801, opened 2023-03-01
Test projects

### LINKED ISSUE #800: Better "Tests" detection
👋 I was trying out fsdocs in repository.  When trying to generate docs for reference information I kept getting:

```
  skipping project 'Project1.fsproj' because it looks like a test project
```

Turns out it's because I'm in a directory that has the word "test" in it. "fsdocstester/src/Project1".

https://github.com/fsprojects/FSharp.Formatting/blob/e665aa95ab897e5e2c543e5847f3820b3e215e92/src/fsdocs-tool/ProjectCracker.fs#L429

I know it's an unlikely case but I'm always the one to hit those 😂 .

nhirschey:
Fixes #800. Mostly relies on projInfos now to exclude test projects, but for building FSharp.Formatting docs correctly we need custom code to exclude this library's own test projects from the ApiDocs.

I exclude FSharp.Formatting test projects using
```fsharp
if s.Contains("tests\\FSharp") then
```

We could be more explicit, such as below, but then if a future person adds additional test projects to this library they might forget to explicitly exclude them

```fsharp
if s.Contains("FSharp.ApiDocs.Tests") || s.Contains("FSharp.Formatting.TestHelpers")
```

### DIFF
--- a/src/fsdocs-tool/ProjectCracker.fs
+++ b/src/fsdocs-tool/ProjectCracker.fs
@@ -426,12 +426,12 @@ module Crack =
         let projectFiles =
             projectFiles
             |> List.choose (fun s ->
-                if s.Contains(".Tests") || s.Contains("test") then
-                    printfn "  skipping project '%s' because it looks like a test project" (Path.GetFileName s)
+                if s.Contains("FSharp.ApiDocs.Tests") || s.Contains("FSharp.Formatting.TestHelpers") then
+                    printfn "  skipping project '%s' because it looks like a FSharp.Formatting test project" (Path.GetFileName s)
                     None
                 else
                     Some s)
-
+        
         //printfn "filtered projects = %A" projectFiles
         if projectFiles.Length = 0 && (ignoreProjects |> not) then
             printfn "no project files found, no API docs will be generated"
--- a/src/fsdocs-tool/ProjectCracker.fs
+++ b/src/fsdocs-tool/ProjectCracker.fs
@@ -425,15 +425,16 @@ module Crack =
         //printfn "projects = %A" projectFiles
         let projectFiles =
             projectFiles
-            |> List.choose (fun s ->
-                // Other libraries' test projects will be filtered out by projInfos later on,
-                // but this make sure FSharp.Formatting.ApiDocs test projects get excluded.
-                if s.Contains("tests\\FSharp") then
-                    printfn "  skipping project '%s' because it looks like a test project" (Path.GetFileName s)
+            |> List.filter (fun s ->
+                let isFSharpFormattingTestProject =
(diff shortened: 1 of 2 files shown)


src/fsdocs-tool/ProjectCracker.fs
```diff
@@ -426,8 +426,11 @@ module Crack =
         let projectFiles =
             projectFiles
             |> List.choose (fun s ->
-                if s.Contains(".Tests") || s.Contains("test") then
+                // Other libraries' test projects will be filtered out by projInfos later on,
+                // but this make sure FSharp.Formatting.ApiDocs test projects get excluded.
+                if s.Contains("tests\\FSharp") then
```
nojaf, line 431: Are you sure about this bit?
When I commented out the `|> List.choose` I was able to run all the tests just fine.

(I ran ` dotnet run --project .\build\build.fsproj -- -t Tests` at the root)

nhirschey: The tests run fine, the problem is that FSharp.Formatting tests run on regular (non-test) projects such as this one: https://github.com/fsprojects/FSharp.Formatting/blob/main/tests/FSharp.ApiDocs.Tests/files/crefLib1/crefLib1.fsproj

crefLib1 itself is not a test project, so if we do not exclude it explicitly somehow then crefLib1 will be added to FSharp.Formatting's api reference here: https://fsprojects.github.io/FSharp.Formatting/reference/index.html

nojaf: Ah ok, I see. Thanks for clarifying that.
In that case, I would go for a slightly more strict approach like:
```fsharp
            let fsharpFormattingTestProjectMarker = $"FSharp.ApiDocs.Tests{Path.DirectorySeparatorChar}files"

            projectFiles
            |> List.filter (fun s ->
                let isTestProject = s.Contains fsharpFormattingTestProjectMarker

                if isTestProject then
                    printfn
                        $"  skipping project '%s{Path.GetFileName s}' because it is a project part of the FSharp.Formatting test suite."

                not isTestProject)
```

To really put emphasis on those selected projects that we are excluding.
The `Path.DirectorySeparatorChar` would make it work cross-platform, I'm not sure the current solution would.

What do you think?

nhirschey: Sure, I just pushed a more strict version based on your suggestion plus the required filtering out of `tests/FSharp.Formatting.TestHelpers/FSharp.Formatting.TestHelpers.fsproj`. The test helpers project will otherwise show up in FSharp.Formatting API docs.

nojaf (approved):
Sorry, I forgot about this 🙈.
Thank you for this PR!

Issue #737, opened 2022-01-27
scripts do not seem to be evaluated in isolation


kMutagene:
I suspect this is an issue that is the root cause of many bugs that we are experiencing currently. 
Most of this seems to be caused in conjunction with using `#r "nuget: ..."` references.

A few observations that lead me to this conclusion:

- referencing different versions of a library in different scripts can cause evaluations to fail, while using the same version everywhere works. There seems to be an issue if there are transitive dependencies in the used libraries. (repro coming soon)
- In some cases, values spill over from one script to another. This happens when different values are bound to the same name in different scripts. [Here is a repro](https://github.com/kMutagene/fsdocs-binding-bug-repro) where the same chart is rendered 2 times although it should be completely different in the second script.
- the order in which the scripts are rendered has an influence on which evaluations fail or not. There can be for example one problematic script that will cause all consecutively rendered htmls to have `no value returned by any evaluator`, while removing it will lead to all those files being rendered correctly.
- These issues are almost always happening in blog-style repos, where scripts are heavily using `#r "nuget: ..."` references, while documentation for libraries seem to work fine most of the time, because they usually reference a single compiled binary in all doc scripts.

This seems to be the underlying issue of almost all problems i have had with fsdocs, but it is very hard to pin down to single lines in scripts. I will update this issue with additional insights.

Here are some issues that i also think are caused by this:
- https://github.com/plotly/Plotly.NET/issues/262
- https://github.com/fsprojects/FSharp.Formatting/issues/635 (A missing Newtonsoft.Json reference was causing problems across scripts there)

dsyme:
Hmmm yes, this can easily cause a load of issues

kMutagene:
So i guess this is the problem? https://github.com/fsprojects/FSharp.Formatting/blob/2c93118a9e3d2491803988ab32d752b1210c6f31/src/FSharp.Formatting.Literate/Evaluator.fs#L198-L203

I tried to look a bit into Evaluator.fs but i have to admit that i do not understand that much besides the comments - but this seems to come from a time where we did not have `dotnet fsi` right? It might be a really naive question, but can't we just start `dotnet fsi script.fsx` as a subprocess of the tool and collect the stdout of that?

dsyme:
[...]

PR #807, opened 2023-03-08
One fsiEvaluator per docs file

nhirschey:
I am pretty sure this fixes https://github.com/fsprojects/FSharp.Formatting/issues/737. Instead of creating one `fsiEvaluator` per docs build, I moved `fsiEvaluator` creation inside the `processFile` function. Thus, one evaluator per file.

Anecdotally, I have a site with lots of docs files that has been stuck on fsdocs version 15.0.0. Any version $\ge15.0.1$ caused the fsdocs build process to fail with a stack overflow that seemed to be related to collecting script parsing/fsc results in a list. With this PR, the site finally builds on a recent fsdocs version.


nojaf (approved):
Hi, I think this is ok to have.
Thanks for looking into this!

Issue #669, opened 2021-04-22
Flag --noapidocs does not prevent project cracking


Joott:
When running the command `dotnet fsdocs watch --eval --noapidocs` in a repository that contains .fsproj files, the tool still attempts to crack the projects for API docs. In a repository that contains no library projects or no projects with `<GenerateDocumentationFile>true</GenerateDocumentationFile>` set, this results in every project being skipped which leads to an error.

![image](https://user-images.githubusercontent.com/39764934/115751614-a47e5a80-a399-11eb-912e-bed46a31cbba.png)

The docs in the `docs` folder could theoretically still be built, since the .fsproj files are not needed. Wouldn't it make more sense to skip the cracking of the projects as well instead of only the writing of the API docs when the flag `--noapidocs` is set?

dsyme:
> In a repository that contains no library projects or no projects with `<GenerateDocumentationFile>true</GenerateDocumentationFile>` set, this results in every project being skipped which leads to an error.

This seems wrong, yes, even without `--noapidocs`

Please take a look around here if you'd like to submit a fix https://github.com/fsprojects/FSharp.Formatting/blob/master/src/FSharp.Formatting.CommandTool/ProjectCracker.fs#L347

Joott:
I made a pull request (#670), which changes the error to a warning and removes the exit function. This way, the building of the docs will not fail when no projects could be cracked, with or without `--noapidocs` set. With `--strict` set it fails in either case.
I also have a proposal for an additional solution (https://github.com/Joott/FSharp.Formatting/commit/f912d5f2e3327055e55632887884a1cbbeba34d5) to this problem: The introduction of the `--ignoreprojects` flag. When this flag is set, the list of project files is set to an empty list, which skips all the steps of project cracking. Additionally, it only fails for no cracked projects with `--strict` when `--ignoreprojects` is false, so this option could be used together with `--strict`.

dsyme:
I merged #670.  

> I also have a proposal for an additional solution (Joott@f912d5f) to this problem: The introduction of the --ignoreprojects flag. When this flag is set, the list of project files is set to an empty list, which skips all the steps of project cracking. Additionally, it only fails for no cracked projects with --strict when --ignoreprojects is false, so this option could be used together with --strict.

This is also ok if you'd like to submit that

Joott:
Thank you! I created a pull request which introduces the --ignoreprojects flag.

dsyme:
Thanks for the contribution!

PR #808, opened 2023-03-16
Add `--ignoreprojects` option to CLI docs

brianrourkeboll:
- Add `--ignoreprojects` option to CLI docs.

The `--ignoreprojects` option was added in #675 to address #669, but it never got added to the [documentation](https://github.com/fsprojects/FSharp.Formatting/blob/64bedb0f9e27e05d41c32f96711803b3b94c2d3e/docs/commandline.md?plain=1#L22-L41)—I only found the option via the issue :). It's a pretty useful option if you have a monorepo with many projects and you don't want to go through contortions to stop `dotnet fsdocs` from trying to crack them all on every invocation.

As an aside, would it ever be worthwhile to generate the CLI options docs from [the source code](https://github.com/fsprojects/FSharp.Formatting/blob/64bedb0f9e27e05d41c32f96711803b3b94c2d3e/src/fsdocs-tool/BuildCommand.fs#L1082-L1172), at least in part, so that they stay synchronized?


nojaf (approved):

nojaf:
Thank you for this PR!

> As an aside, would it ever be worthwhile to generate the CLI options docs from [the source code](https://github.com/fsprojects/FSharp.Formatting/blob/64bedb0f9e27e05d41c32f96711803b3b94c2d3e/src/fsdocs-tool/BuildCommand.fs#L1082-L1172), at least in part, so that they stay synchronized?

Good idea, I would accept a PR for that.

PR #812, opened 2023-03-27
Update ipynb metadata so that it works with new versions of polyglot notebook in VsCode

### LINKED ISSUE #809: fsdocs does not generate correct code cell metadata for .ipynb
Dotnet interactive expects specific metadata in an F# notebook. This has evolved over time, and we need to update fsdocs ipynb output to match what they currently expect. See for example https://github.com/dotnet/interactive/issues/2685

I reported the issue in the dotnet/interactive repo a few weeks ago, and they fixed it. But now it's broken again as of ths week's dotnet interactive. We should really make fsdocs output the metadata that dotnet interactive expects to help avoid breakages like this.

When the metadata is not correct, sometimes polyglot notebooks think an F# cell is C#. With correct metadata, this will not happen.

nhirschey:
This fixes #809 by adding additional metadata that is now expected by the VS Code polyglot notebooks extension (formerly known as dotnet interactive extension). See the issue for further info.

I've tested this in VS Code and VS Code insiders.

I added F# metadata for code cells to prevent them being mistaken for C# cells. For notebook metadata I made F# the default notebook language. These issues were not present in earlier versions of polyglot notebooks, but now there's problems for F# notebooks if this metadata is not set.

(sorry for the commit history; something weird happened when I tried to update from the main branch).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+## 17.4.1
+
+* Update ipynb output metadata [#809](https://github.com/fsprojects/FSharp.Formatting/issues/809)
+
 ## 17.4.0
 
 * One FSI evaluator per docs file [#737](https://github.com/fsprojects/FSharp.Formatting/issues/737)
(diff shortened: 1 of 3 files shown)


nojaf:
I would like to merge this one before any of the other PRs, however, I don't feel qualified to review this one as I don't know anything about notebooks.

@nhirschey if you feel confident about this, I would be ok to merge this anyway.

PR #813, opened 2023-03-28
Update to Fantomas v6.

nojaf:
See [blog post](https://blog.nojaf.com/).
There is little impact on this repository.
Most changes you see are actually coming from `5.1` and `5.2`.


nhirschey:
Looks good to me. Mostly noticed fewer spaces and the new version doing a better job condensing code to a single line where possible.

PR #796, opened 2023-02-16
Update FSharp.Compiler.Service to 43.7.200

nojaf:
A naive attempt to update FCS.

PR #816, opened 2023-04-05
Pass `--multiemit-` as default option for FsiEvaluator.

nojaf:
After updating to v18 in Fantomas, we noticed that Linux is no longer processing `(*** include-output ***)`.

When locally (in WSL) adding `--multiemit-` to `FsiEvaluator` is started working again.
I propose we ship this out of the box.

Thoughts @baronfel?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+## 18.0.1
+
+* Pass `--multiemit-` as default option for `FsiEvaluator`.
+
 ## 18.0.0
 
 * Update FCS to 43.7.200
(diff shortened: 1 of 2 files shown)


dsyme:
It would be good to understand what the problem is.  Any specifics?

It's ok to turn off multiemit (it's primarily for debug symbols) but I'd like to understand better what's failing.

nojaf:
@dsyme it first surfaced during [the upgrade to FCS 43](https://github.com/fsprojects/FSharp.Formatting/pull/796).

Some `eval` calls would stop working after the second invocation.
It was a bit icky to troubleshoot, it would happen when you run two unit tests (from `FSharp.Literate.Tests.Eval`) sequentially.
Running either test individually would work, but the last test would fail after evaluation.

nojaf:
We should definitely investigate this behaviour, but I'm going to merge this in for the time being as it does break certain use cases.

nhirschey:
This didn’t bump the release. The nuget version is still 18.1.0

https://www.nuget.org/packages/FSharp.Formatting

PR #820, opened 2023-04-14
Replace Paket with NuGet

nojaf:
Similar to https://github.com/fsprojects/FSharp.Formatting/pull/817, I think that using Paket for this project is unnecessary.

With the availability of lock files and central package version management in NuGet, which is now part of the SDK, I feel that this approach is more practical.

Although this PR might spark some disagreement, I am confident that it will make it easier for new users to get involved.


smoothdeveloper:
@nojaf, I am curious what exactly makes paket less practical? As far as I can see, updating the dependencies in this PR will require to make sure the packages.lock.json across the source tree are also updated, while with paket, all is in paket.lock, also the definition of dependencies doesn't seem to be simpler to me.

I personally doubt the using paket is a blocker for contributions.

nojaf:
- NuGet is the standard package manager for dotnet, and there is in general more familiarity and community support.
- NuGet is guaranteed to work after every new SDK release. If you encounter issues with Paket and it doesn't work for your specific needs, it can be a frustrating experience.
- Better tooling support: NuGet has better tooling support than Paket, which means that it is easier to integrate with other development tools and platforms.

dsyme (approved):

PR #835, opened 2023-08-24
Silence warnings about unused values during compilation

dawedawe:
Not really a big issue, but during compilation there are several compiler warnings flying by about unused values.
This PR silences these warnings.


nojaf:
Hmm, I think it might be slightly easier to go with a `nowarn` flag somewhere.
Not hellbent on it though.

dawedawe:
Yeah, good call.

nojaf (approved):
Thank you!

Issue #736, opened 2022-01-20
New lines in API reference docs are not included 


kMutagene:
New lines seem to be omitted from XML docs in the created API documentation. See below screenshot for a comparison of the input (left) and the generated output (right)

This makes large structured docs hard to read.

![image](https://user-images.githubusercontent.com/21338071/150303827-1dfc2102-f67e-4bfe-96fc-98eac4bc1737.png)

rdipardo:
Quick fix: assuming none of the docs contain explicit XML markup (`<summary>...</summary>`, etc.),
just use the `--mdcomments` [command line option] to render everything as Markdown:

[command line option]: https://fsprojects.github.io/FSharp.Formatting/commandline.html#The-build-command



![mdcomments_example](https://user-images.githubusercontent.com/59004801/150771913-d9a867ed-c904-4856-8d29-c22767ff2eaa.png)

dsyme:
Good answer 👍

PR #843, opened 2023-09-08
Add pre tag for comment summary.

nojaf:
Fixes https://github.com/fsprojects/FSharp.Formatting/issues/736.
@kMutagene would you mind giving this a spin on your local codebase?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+## 19.0.1
+
+* Fix New lines in API reference docs are not included [#736](https://github.com/fsprojects/FSharp.Formatting/issues/736)
+
 ## 19.0.0
 
 * Update FCS to 43.7.400
--- a/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
@@ -185,7 +185,7 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
                                       yield! copyXmlSigIconForSymbolMarkdown m.Symbol
                                       yield! copyXmlSigIconForSymbol m.Symbol
                                       yield! sourceLink m.SourceLocation
-                                      p [ Class "fsdocs-summary" ] [ embed m.Comment.Summary ]
+                                      p [ Class "fsdocs-summary" ] [ pre [] [ embed m.Comment.Summary ] ]
                                   ]
 
                               let dtls =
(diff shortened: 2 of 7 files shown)


dsyme:
Hmmm is using `pre` really right?  Any chance of a before/after visual on the generated docs?    If I look at the docs for, say, Diffsharp, it would look wrong to use pre-formatted text I'm sure:

image


It looks like the original request was in the context of summary text written as markdown. In theory, formatted text should be written using an explicity `<summary>` and `<code></code>` and `<para></para>` sections etc.

I agree with your original analysis that if people are writing docs as markdown, they should use the original flag provided for that.   Perhaps you could provide a flag for the behaviour implemented here?  But it shouldn't be the default

dsyme:
Also, the Summary will be formatted HTML at this point, and include `<b>...</b>` and other elements, converted from the XML doc via https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/GenerateModel.fs#L1827

Arguably what should be happening is the F# compiler supports a `--markdowndocs` option, which interprets `/// text` as a limited subset of markdown and emits an XML doc that corresponds to the intent of the markdown, or else emits a single special `<markdown>...</markdown>` section with the raw markdown (not totally sure which is better - the former would mean consuming F# tooling doesn't need to be updated - the latter would require updates to all consuming tooling)

nojaf:
Hmm, I still think this makes sense, to be honest.
Consider:

```fsharp
/// To use this from C#
///
///     [lang=csharp]
///     var a = new MyClass()
///
/// To use this from F#
///
///     let a = FsLib.MyClass()
///
type MyClass() =
    member x.Nothing = 0
```

This now (on main) generates:
```html
  <div class="fsdocs-xmldoc">
    <div>
      <p class="fsdocs-summary">
        
 To use this from C#

     [lang=csharp]
     var a = new MyClass()

 To use this from F#

     let a = FsLib.MyClass()


      </p>
    </div>
  </div>
```
Wrapping this in a summary tag does not affect the output:
```fsharp
/// <summary>
/// To use this from C#
///
///     [lang=csharp]
///     var a = new MyClass()
///
/// To use this from F#
///
///     let a = FsLib.MyClass()
/// </summary>
type MyClass() =
    member x.Nothing = 0
```

same HTML:
```fsharp
  <div class="fsdocs-xmldoc">
    <div>
      <p class="fsdocs-summary">
        
 To use this from C#

     [lang=csharp]
     var a = new MyClass()

 To use this from F#

     let a = FsLib.MyClass()
 
      </p>
    </div>
  </div>
```

The diffsharp example you mention is a little different:
https://github.com/DiffSharp/DiffSharp/blob/29af24ad69a96aace8a49420996861b2a3ca0fa9/src/DiffSharp.Core/DiffSharp.fs#L977-L979

The summary/remark gives:
```fsharp
/// <summary>
/// To use this from C#
///
///     [lang=csharp]
///     var a = new MyClass()
/// </summary>
/// <code>
/// To use this from F#
///
///     let a = FsLib.MyClass()
/// </code>
/// <remarks>Yes, ok new paragraph</remarks>
type MyClass() =
    member x.Nothing = 0
```

```html
  <div class="fsdocs-xmldoc">
    <div>
      <p class="fsdocs-summary">
        
 To use this from C#

     [lang=csharp]
     var a = new MyClass()
 
      </p>
    </div>
    <p class="fsdocs-remarks">
      Yes, ok new paragraph
    </p>
  </div>
```

Which still doesn't take any newlines into account for the individual paragraph element.

nojaf:
This was indeed not the way to go.

PR #781, opened 2022-11-18
Option to have dark code schema ?

Thorium:
I was using this project to replace generate.fsx docs for [SQLProvider](https://fsprojects.github.io/SQLProvider/). Worked pretty well, but **the default colour schema is a bit "from 70s sofa"** theme and I missed the dark more-VS-style-editor theme. So I added this theme file, it's not perfect, but feel free to improve.

I don't know how this project works, but I just run `--parameters fsdocs-theme darkcode` and even though all the docs says not supported, when I copied the file to my output content directory, behold!

The darkcode, this PR:
image

...versus the default scheme:
image

But as this was initial version, I definitely want to keep the original too, because it might feel comfy for some.

PR #844, opened 2023-09-11
Update to bootstrap 5.

nojaf:
This PR gives the default template a facelift with Bootstrap 5. This update makes it smoother to switch between light and dark modes, thanks to Bootstrap [v5.3](https://blog.getbootstrap.com/#dark-mode).

I also borrowed some ideas from Jimmy, which you can check out in this [comment](https://github.com/fsprojects/FSharp.Formatting/pull/781#issuecomment-1453667661).

Take a peek at the new look: DarkLightFSharpFormatting

Now, I'm wondering if it's time to rethink our reliance on Bootstrap. Maybe it's more convenient to start fresh with a custom CSS theme using CSS variables for just about everything. This way, we can effortlessly handle light and dark modes from the get-go and maintain it easily. If we decide to tweak the default theme, users can do so by overriding a bunch of CSS variables.

Let's chat about this! Tagging some folks who know this repo: @dsyme, @nhirschey, @baronfel, @kMutagene, and @theAngryByrd.

### DIFF
--- a/docs/content/darkLight.js
+++ b/docs/content/darkLight.js
@@ -1,0 +1,84 @@
+import {LitElement, html, css} from 'https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js';
+import 'https://code.iconify.design/iconify-icon/1.0.7/iconify-icon.min.js';
+
+export class DarkLightToggle extends LitElement {
+    static properties = {
+        _theme: 'light'
+    }
+
+    constructor() {
+        super();
+        this.theme = localStorage.getItem("theme");
+        if (this.theme === "dark") {
+            document.documentElement.setAttribute("data-bs-theme", "dark");
+            this._theme = "dark";
+        }
+    }
+
+    toggleTheme() {
+        const current = document.documentElement.getAttribute("data-bs-theme");
+        if (current === "dark") {
+            document.documentElement.setAttribute("data-bs-theme", "light");
+            localStorage.setItem("theme", "light");
+            this._theme = "light";
+        } else {
+            document.documentElement.setAttribute("data-bs-theme", "dark");
+            localStorage.setItem("theme", "dark");
+            this._theme = "dark";
+        }
+    }
+
+    static styles = css`
+      div {
+        margin: 1rem 0;
+      }
+
+      input[type=checkbox] {
+        opacity: 0;
+        position: absolute;
+        margin: 0;
+      }
+
+      label {
+        background-color: rgb(34, 34, 34);
+        width: 45px;
+        height: 20px;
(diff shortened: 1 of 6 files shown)


kMutagene:
The dark theme looks really good!

> Maybe it's more convenient to start fresh with a custom CSS theme using CSS variables for just about everything

I actually started something similar a while back on https://github.com/fslaborg/docs-template. the gist is that i used [Bulma]() instead of bootstrap, which is a css only framework and can be customized via sass and variables.   

If any css variable-based customization would find its way directly into this library, i think that would be the starting point for an actual templating system. Wether you'd want to switch to bulma or do something completely from scratch does not matter too much i think.

nhirschey:
I built a site with the new template and it generally looks good. I have no opinion on css vs. bootstrap.

Regarding the dark theme, I think the blue used for open/let/for/in keywords is too close to the background color. I cannot easily read these keywords in the below dark mode screenshot.

image

docs/content/darkLight.js
```diff
@@ -0,0 +1,84 @@
...
+    constructor() {
+        super();
+        this.theme = localStorage.getItem("theme");
+        if (this.theme === "dark") {
+            document.documentElement.setAttribute("data-bs-theme", "dark");
+            this._theme = "dark";
+        }
+    }
+
+    toggleTheme() {
```
TheAngryByrd, line 18: It's worth considering using [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) so we use whatever someone's systems defaults are.

TheAngryByrd (commented):
❤️ this.  Left something to consider.

docs/content/darkLight.js
TheAngryByrd: image


I know it's not as tidy as a toggle button but I think respecting the user's system theme is more important. Some example code: [code](https://github.com/TheAngryByrd/IcedTasks/blob/cdb90c260daabb499f8a0988046238dc933061a3/docsSrc/content/theme-toggle.js#L17)

dsyme:
I'm ok with this!

nojaf:
Thanks all for the feedback. I'm still on the fence about using bootstrap or not.
I'm currently not using it for the new [Ionide analyzers website](https://ionide.io/FSharp.Analyzers.SDK/) and I think it might be reasonable to rewrite the default theme here from scratch. The CSS variable approach is working out quite well.

It could make things easier to theme, have fewer dependencies and foresee the light/dark from the start.

nojaf:
Closing in favour of https://github.com/fsprojects/FSharp.Formatting/pull/861

Issue #869, opened 2023-11-14
SEO-optimization for new theme


Thorium:
The current documentation pages should win Google search by default, but GitHub and Nuget seems to go before that.
So... The HTML pages could have better parameters in the head of the page, for example:
```html
<head>
<!-- ... current ones plus ... -->
<meta name="description" content="Library used for building custom analyzers for FSAC / F# editors.">
<!-- OGs are for social media sharing previews -->
<meta property="og:site_name" content="FSharp.Analyzers.SDK">
<meta property="og:image" content="http://ionide.io/FSharp.Analyzers.SDK/images/logo.png">
<meta property="og:image:secure_url" content="https://ionide.io/FSharp.Analyzers.SDK/images/logo.png">
<!-- match Twitter option with Some 
<meta name="twitter:card" content="summary_large_image">
<meta name="twitter:site" content="@fsharp_analysers">
<meta name="twitter:image" content="http://ionide.io/FSharp.Analyzers.SDK/images/logo.png">
-->
<!-- This would be for your mother's phone: -->
<!--link rel="apple-touch-icon-precomposed" sizes="144x144" href="http://ionide.com/apple-touch-icon-144-precomposed.png"-->
</head>
```

kMutagene:
If i understand the current system correctly, there is no default `_template.html` file provided by the tool, and users are expected to create their own. I think a `fsdocs init` command that provides a html template that contains the seo tags would be nice, but to make this automated, a lot of parameters have to be supplied by the user.

Example:

This is how i try to do automatic seo tags for each documentation page on [plotly.net](https://plotly.net):

```html
    <meta charset="utf-8">
    <title>{{fsdocs-page-title}}</title>
    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
    <meta name="author" content="{{fsdocs-authors}}">
    <meta name="description" content="Plotly.NET is an Interactive charting library for .NET programming languages. This page contains the documentation for: {{fsdocs-page-title}}">

    <!-- Opengraph properties (https://ogp.me/) -->
    <meta property="og:title" content="Plotly.NET documentation for: {{fsdocs-page-title}}" />
    <meta property="og:url" content="{{root}}{{fsdocs-source-basename}}.html">
    <meta property="og:type" content="website" />
    <meta property="og:site_name" content="Plotly.NET" />
[...]

PR #873, opened 2023-11-17
add minimal seo tags for open graph and twitter cards to default template

kMutagene:
see #869 

This is a basic implementation using the available substitution parameters. There could be more, e.g. displaying a default image etc., but i think for that we would need additional substitutions.

### DIFF
--- a/docs/_template.html
+++ b/docs/_template.html
@@ -7,6 +7,18 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="author" content="{{fsdocs-authors}}">
 
+    <!-- Opengraph properties (https://ogp.me/) -->
+    <meta property="og:site_name" content="{{fsdocs-collection-name"}}">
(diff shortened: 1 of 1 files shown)


Thorium:
Is there something good that could be added to the meta description tag?

nojaf:
Maybe we can use the `<Description>` tag from the cracked projects?

kMutagene:
> Maybe we can use the [Description](https://learn.microsoft.com/en-us/nuget/create-packages/creating-a-package-msbuild#add-an-optional-description-field) tag from the cracked projects?

I was thinking about that, but this has the same problem as `fsdocs-authors` has - if the respective xml tag in the cracked project does not exist, the tag will contain rubbish. However i agree that that tag would be a sensible default when it exists.

nojaf:
We could also fix that part. Having better defaults or use empty strings if not replaced. Or let the tool crash in a strict mode if values are missing.

PR #874, opened 2023-11-17
Ipynb input

### LINKED ISSUE #806: Using polyglot notebooks (dotnet interactive) as input
Emitting notebooks with this tool is a killer feature. Since there is already compatibility with the internal model and the ipynb format, i suggest another notebook-based feature: **using notebooks as input**. I would love to help implement this, but would need some pointers on how to navigate the code base.

This has huge advantages over working with literal scripts: 
- one can work on a single notebook in an editor such as vscode and interactively evaluate it (e.g. check that the markdown renders correctly or inspect plots and other interactive output) without having to run the tool in watch mode (which for example takes more than 10 minutes to start up for a project with many literate scripts such as plotly.net.
- It makes colaborating on docs and blog-post style documents way easier, since it eliminates getting the actual project running with it's intricacies regarding the individual build chains. You just have to commit a single notebook file.

This would be similar to python's [nbconvert](https://nbconvert.readthedocs.io/en/latest/), and i think such a tool is a critical part missing in the .NET notebook landscape. Maybe it would be better to make a standalone tool for this, but that is up for debate, i focused on this repo since it can already generate notebooks.

Note that you can already use nbconvert to convert polyglot notebooks to html (which is what i currently do), but that means you have to install and maintain a python environment instead of being able to use a dotnet tool.

nhirschey:
Fixes #806. This code should be independent of everything else; only gets triggered on .ipynb input files. Could merge this into the modern theme branch for alphas or wait and merge into main after modern theme is complete. 

The general strategy is to convert a notebook to markdown and then pass that markdown through FSharp.Formatting's existing markdown tools.

**Parsing:** This uses a custom parser to read notebooks as json. I tried using .NET Interactive's parser, but it requires mapping their undocumented internal C# classes to F# and at the end of the day it was easier to do the parsing from scratch. 

**Evaluation:** relies on [dotnet-repl](https://github.com/jonsequitur/dotnet-repl), a personal project of someone on the .NET Interactive team. As recently as August they recommended it for commandline execution, so let's use it. The downside is that users must install this extra tool, but I don't think it's a big burden.

One thing @kMutagene, we have to decide what happens when we evaluate a notebook. What I did was have `fsdocs --eval` trigger dotnet-repl on the actual source notebook in the docs folder. This means that evaluating docs will change the content of the source notebooks (for example dotnet-repl puts some additional metadata in cells indicating time of execution). This seemed reasonable for now.

**Testing:** I added tests, but one thing I couldn't test well was dotnet-repl evaluation. I couldn't get the test project to evaluate the notebooks using dotnet-repl. Maybe test projects cannot call external dotnet tools? I don't know. I left the test in there but it's commented out.


src/FSharp.Formatting.Literate/ParsePynb.fs
```diff
@@ -0,0 +1,162 @@
...
+                | Some outputs ->
+                    let outputsString = outputs |> String.concat "\n"
+                    sprintf $"{codeBlock}\n{outputsString}"
+
+    module Output =
+        let (|TextHtml|_|) (x: JsonElement) =
+            match x.TryGetProperty("text/html") with
+            | true, html ->
+                let html = html.EnumerateArray() |> Seq.map (fun x -> x.GetString()) |> String.concat "\n"
+                Some("""<p>""" + html + """</p>""")
```
nojaf, line 34: Nitpick: use interpolated string instead of `+`.

src/FSharp.Formatting.Literate/ParsePynb.fs
```diff
@@ -0,0 +1,162 @@
...
+            | _ -> None
+
+        let (|TextPlain|_|) (x: JsonElement) =
+            match x.TryGetProperty("text/plain") with
+            | true, text ->
+                let text = text.EnumerateArray() |> Seq.map (fun x -> x.GetString()) |> String.concat ""
+
+                Some(
+                    """<table class="pre"><tbody><tr><td><pre><code>"""
```
nojaf, line 43: Nitpick: use interpolated string instead of `+`.

src/FSharp.Formatting.Literate/ParsePynb.fs
```diff
@@ -0,0 +1,162 @@
...
+                    + """</code></pre></td></tr></tbody></table>"""
+                )
+            | _ -> None
+
+        let (|DisplayData|_|) (x: JsonElement) =
+            if x.GetProperty("output_type").GetString() = "display_data" then
+                match x.GetProperty("data") with
+                | TextHtml html -> html
+                | TextPlain text -> text
+                | s -> failwith $"unknown ouptut {s}"
```
nojaf, line 54:
```suggestion
                | s -> failwith $"unknown output {s}"
```

src/FSharp.Formatting.Literate/ParsePynb.fs
```diff
@@ -0,0 +1,162 @@
...
+
+                Some(
+                    """<table class="pre"><tbody><tr><td><pre><code>"""
+                    + text
+                    + """</code></pre></td></tr></tbody></table>"""
+                )
+            | _ -> None
+
+        let (|DisplayData|_|) (x: JsonElement) =
+            if x.GetProperty("output_type").GetString() = "display_data" then
```
nojaf, line 50: Would it not make sense to be a little bit more defensive in these active patterns?
Use `x.TryGetProperty` to avoid a null pointer if `"output_type"` is not present?

src/FSharp.Formatting.Literate/ParsePynb.fs
```diff
@@ -0,0 +1,162 @@
...
+
+        let (|Stream|_|) (x: JsonElement) =
+            if x.GetProperty("output_type").GetString() = "stream" then
+                let text =
+                    x.GetProperty("text").EnumerateArray()
+                    |> Seq.map (fun x -> x.GetString())
+                    |> String.concat ""
+
+                Some(
+                    """<table class="pre"><tbody><tr><td><pre><code>"""
```
nojaf, line 67: Nit: interpolated string

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -186,6 +187,8 @@ type internal DocContent
 
               let isMd = inputFileFullPath.EndsWith(".md", true, CultureInfo.InvariantCulture)
 
+              let isPynb = inputFileFullPath.EndsWith(".ipynb", true, CultureInfo.InvariantCulture)
```
nojaf, line 190:
```suggestion
              let isPynb = inputFileFullPath.EndsWith(".ipynb", StringComparison.Ordinal)
```

You actually want to compare without any culture at all.
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -335,6 +338,61 @@ type internal DocContent
                                        template,
                                        outputFileFullPath
                                    )))
+                      elif isPynb then
+                          printfn "  preparing %s --> %s" inputFileFullPath outputFileRelativeToRoot
+
+                          let evaluateNotebook ipynbFile =
+                              let psi =
+                                  ProcessStartInfo(
```
nojaf, line 346: This will fail when the user doesn't have `dotnet repl` available.
image

I think we want to check if we can use this by running `dotnet repl --version` first.
And, in any case, we probably want to wrap running external processes in a `try/with` block.

tests/FSharp.Literate.Tests/DocContentTests.fs
```diff
@@ -225,3 +254,38 @@ let ``Parses frontmatter correctly `` () =
     twoTowersHtml |> shouldContainText "<a href=\"fellowship.html\">Previous</a>"
     twoTowersHtml |> shouldContainText "<a href=\"return.html\">Next</a>"
     returnHtml |> shouldContainText "<a href=\"two-tower.html\">Previous</a>"
+
+
+(* Cannot get this test to evaluate the notebook
```
nojaf, line 259: Why not? What goes wrong?

nojaf (commented):
This is a great work! I have some nitpicks and some places I would code a little more defensive. But overall this is solid!

src/fsdocs-tool/BuildCommand.fs
nhirschey: Changed all 3 to StringComparison.OrdinalIgnoreCase. The old version ignored case so that `file.md` and `file.MD` are both treated as markdown files.

tests/FSharp.Literate.Tests/DocContentTests.fs
nhirschey: I have `evaluate = true` set for DocContent, which should trigger the external call to `dotnet repl` that runs the notebook and produces outputs for the cell . But for some reason  the code calling out to `dotnet repl` isn't getting triggered inside this test. `dotnet fsdocs --eval` triggers notebook evaluation in a docs folder, but I have not been able to get it to work inside this test.

nojaf: Are you sure? This test is working for me in `WSL`.

nhirschey:
Sounds good, I'll bump the alpha release when I merge.

Thanks for the code review and all your work on the modern theme. You've done incredible work the past month!

nhirschey:
@nojaf, just to double check that I don't mess up any of your in-progress work. Are you ok with this being merged into main now? I'm done and I've updated release notes to bump the release

nojaf:
> You've done incredible work the past month!

Why thank you 😊

> Are you ok with this being merged into main now?

Yes, go for it!

jonsequitur:
> Parsing: This uses a custom parser to read notebooks as json. I tried using .NET Interactive's parser, but it requires mapping their undocumented internal C# classes to F# and at the end of the day it was easier to do the parsing from scratch.

If there are changes you'd suggest making to Microsoft.DotNet.Interactive.Documents`, please let us know. Ideally you'd be able to use this effectively from F# without needing to duplicate the effort.

jonsequitur:
> Evaluation: relies on [dotnet-repl](https://github.com/jonsequitur/dotnet-repl), a personal project of someone on the .NET Interactive team. As recently as August they recommended it for commandline execution, so let's use it. The downside is that users must install this extra tool, but I don't think it's a big burden.

I'm the owner of `dotnet-repl`. It's intended as a prototype for command line functionality that could find its way into the core of .NET Interactive. I'm glad you're finding it useful. I'm also more than happy to work with people on contributions to it.

PR #876, opened 2023-11-20
Use analyzers

nojaf:
See https://ionide.io/FSharp.Analyzers.SDK/


nojaf:
@dawedawe it might be interesting to take a look at the exceptions produced during the `TAST` walking.

dawedawe:
> @dawedawe it might be interesting to take a look at the exceptions produced during the `TAST` walking.

Yes, I'll take a look.

PR #901, opened 2024-02-08
Avoid theme flicker in dark mode.

nojaf:
Fix found by @TheAngryByrd to avoid the flickering of the dark theme.


nhirschey:
Wow, nice. The flicker was a bit annoying.

nojaf:
Yeah, I don't use dark mode so I wasn't overly annoyed by it 😅

PR #902, opened 2024-02-14
Stable v20

nojaf:
Releasing v20 as stable!

//cc @nhirschey @kMutagene


nhirschey:
Wonderful achievement @nojaf! Thanks for all your work bringing this modernization to fruition. I know it was quite a bit of work, but it turned out really nicely.

kMutagene:
Thank you both so much for this, notebooks input combined with the new look makes this tool so much better. I hope I can come around contributing soon again

PR #904, opened 2024-02-28
Make example command in docs/literate-notebook.ipynb verbose

johnW-ret:
Changes use of `dotnet fsdocs --eval` to `dotnet fsdocs [build|watch] --eval`. I should have known but it cost me a bit of time as a beginner so I think the verbosity would help 🙂.

### DIFF
--- a/docs/literate-notebook.ipynb
+++ b/docs/literate-notebook.ipynb
@@ -2,9 +2,10 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "91a9b093",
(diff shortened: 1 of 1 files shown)


nojaf:
Hi @johnW-ret, thanks for this PR. I'm going ask @nhirschey to review this one.
It looks fine to me but I don't have enough understanding about notebooks.

nhirschey:
Thank you @johnW-ret, I agree this is better. Sorry for the confusion!

PR #906, opened 2024-03-15
Minor code clean-ups and performance improvements

Thorium:
Code clean-up with minor performance optimizations:

- isNull instead of "= null": https://latkin.org/blog/2015/05/18/null-checking-considerations-in-f-its-harder-than-you-think/
- TryGetValue instead of double-lookups of dictionaries
- Types and active patterns that are not containing any data can be structs to reduce object capsuling and memory footprint
- inline functions preferably internal to not leak outside of the project

No functional changes.

### DIFF
--- a/src/Common/StringParsing.fs
+++ b/src/Common/StringParsing.fs
@@ -15,8 +15,9 @@ open FSharp.Formatting.Markdown
 
 module String =
     /// Matches when a string is a whitespace or null
+    [<return: Struct>]
     let (|WhiteSpace|_|) (s) =
-        if String.IsNullOrWhiteSpace(s) then Some() else None
+        if String.IsNullOrWhiteSpace(s) then ValueSome() else ValueNone
 
     /// Returns a string trimmed from both start and end
     let (|TrimBoth|) (text: string) = text.Trim()
@@ -117,15 +118,17 @@ module String =
 
 module StringPosition =
     /// Matches when a string is a whitespace or null
+    [<return:Struct>]
     let (|WhiteSpace|_|) (s, _n: MarkdownRange) =
-        if String.IsNullOrWhiteSpace(s) then Some() else None
+        if String.IsNullOrWhiteSpace(s) then ValueSome() else ValueNone
 
     /// Matches when a string does starts with non-whitespace
+    [<return:Struct>]
     let (|Unindented|_|) (s: string, _n: MarkdownRange) =
         if not (String.IsNullOrWhiteSpace(s)) && s.TrimStart() = s then
-            Some()
+            ValueSome()
         else
-            None
+            ValueNone
 
     /// Returns a string trimmed from both start and end
     let (|TrimBoth|) (text: string, n: MarkdownRange) =
@@ -174,11 +177,12 @@ module StringPosition =
              StartColumn = n.StartColumn + text.Length - trimmed.Length })
 
     /// Matches when a string starts with any of the specified sub-strings
+    [<return:Struct>]
     let (|StartsWithAny|_|) (starts: string seq) (text: string, _n: MarkdownRange) =
         if starts |> Seq.exists (fun s -> text.StartsWith(s, StringComparison.Ordinal)) then
-            Some()
+            ValueSome()
         else
-            None
+            ValueNone
 
     /// Matches when a string starts with the specified sub-string
     let (|StartsWith|_|) (start: string) (text: string, n: MarkdownRange) =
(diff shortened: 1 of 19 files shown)


src/Common/StringParsing.fs
```diff
@@ -15,8 +15,12 @@ open FSharp.Formatting.Markdown
 
 module String =
     /// Matches when a string is a whitespace or null
+    [<return: Struct>]
```
nojaf, line 18: Do you have any idea under which conditions you would start using this?

- What is the minimum language version?
- Always for unit or can this be used for other return types as well?

src/Common/StringParsing.fs
```diff
@@ -297,15 +307,16 @@ module StringPosition =
...
-        | StartsWithRepeated repeated (_n, (v, _)) when (String.IsNullOrWhiteSpace v) -> Some()
-        | _ -> None
+        | StartsWithRepeated repeated (_n, (v, _)) when (String.IsNullOrWhiteSpace v) -> ValueSome()
+        | _ -> ValueNone
 
 module List =
     /// Matches a list if it starts with a sub-list that is delimited
     /// using the specified delimiters. Returns a wrapped list and the rest.
-    let inline (|DelimitedWith|_|) startl endl input =
+    let inline internal (|DelimitedWith|_|) startl endl input =
```
nojaf, line 319: So `internal` because this is used in other files in the project?
Otherwise `private` would also work?

src/FSharp.Formatting.Common/YaafFSharpScripting.fs
```diff
@@ -464,26 +463,29 @@ module internal ArgParser =
...
-                | _ -> None
-        | _ -> None
+                | "+" -> ValueSome true
+                | "-" -> ValueSome false
+                | _ -> ValueNone
+        | _ -> ValueNone
 
 open ArgParser
 
+[<Struct>]
```
nojaf, line 481: The heuristic would be here to add `[<Struct>]` when the DU only has cases without any fields?

nojaf (approved):
Hi @Thorium, this is quite insightful.

It sparks some interest to detect these things via analyzers.
Before I dwindle in that territory, did you detect these manually? Or use some tooling?

Last question: is any of this noticeable? We don't have a benchmark so I'm not asking you to add that but did you run this before and after? Just curious.

src/Common/StringParsing.fs
Thorium: Struct constructs came with FSharp.Core 6.0.1 so that's the minimum version. 
Works with other return types too.

src/Common/StringParsing.fs
Thorium: I just want to limit the scope to this solution: As you know inline makes compiler slower. So if you have hundreds of public inline functions in your solution and someone refers your project via Nuget package, they are going to pay the price with their compilation too, because those inline-functions come available for them too, even when they probably are not aware of them and probably will never need them.

src/FSharp.Formatting.Common/YaafFSharpScripting.fs
Thorium: Yes, Enums and similar are struct per default, so if you do this:
```fsharp
type MyE =
| A = 1
| B = 2
```
it's fine as is. But if you do this:
```
type MyE =
| A 
| B 
``
...it's not a struct. So basically it'll generate a class of MyE containing a class of A and class of B and an instance of either.

**Comment by Thorium** [comment]
General guidelines about structs (and I don't have good references to point here):

*Keep your structs small.*

By small I mean a bool or guid or int or int64 or another struct, but not bunch of other classes or interface-implementations of any kind or big lists or anything like that. 

Struct is read to memory at once and kept there, so it's not a pointer to somewhere else, and it makes the processing fast and footprint small. The limitation is function parameters, structs are copied and not sent as pointers, which means if you have big structs, then function calls go slow. Which you can avoid by making your structs as ByRef and so on, but then they are mutable and it goes as hack-hack rather than proper enterprise code. But that's why e.g. System.Text.Json is so fast, they use big structs byref and trust to take 100% out of your single core.

If you have a simple object having a few non-class fields, then it's better to be struct. If you have a 1000s of these small objects it will really improve the performance and memory footprint.

Option class is not a struct, but there is a struct-corresponding ValueOption. ValueOption has IsSome and IsNone and typically the usage with match-clause will just work. And there is ValueOption.bind and so-on like Option.bind you've used to. But if you use lot of existing F# functions like List.tryPick they work with Some x and not ValueSome x, and there I'd go for usability over performance.

Tuple class is not a struct but, but there is a ValueTuple which is used instead of `(a,b)` like this: `struct(a,b)`. That's a bit inconvenient to use.

So this is still good:

```fsharp
[<Struct>]
type MyType =
| MyOp of int
| MyOp2 of bool ValueOption 
| MyOp3 of struct(bool*bool) 
```

They also say you should performance-test, to verify your changes, because many times it depends on use-case if a harmless looking `[<Struct>] type MyType = MyOp of int array` is better to be a struct or not.

So for conservative analysis, I'd go for only non-data-carrying and/or only plain boolean/int/guid/struct-field types to be recommended to be structs.

nojaf:
Thanks for all the insight!

> Last question: is any of this noticeable? We don't have a benchmark so I'm not asking you to add that but did you run this before and after? Just curious.

src/FSharp.Formatting.Common/YaafFSharpScripting.fs
```diff
@@ -464,26 +463,29 @@ module internal ArgParser =
         else
             None
 
+    [<return: Struct>]
```
Thorium, line 466: @nojaf here is an example where  there is a Struct boolean. As boolean is a tiny field, there is no point of wrapping it to inside a separate class, like the original code did.

Thorium:
The comparisons are probably microseconds, so I expect it's not human noticeable in a single run of generating the docs, but if you have build-pipeline over every commit queuing generating docs, then even small changes can affect over time, to free up some CI workers faster.

src/Common/StringParsing.fs
Thorium: This is my reasoning only, and I might be wrong:
If do code like this: `myInt |> ` and now you ask your F# IDE to auto-complete, then when it knows the type of "myInt" then it can give proper function options: Those who take int in this case. 

But in case of inline functions, they use statically resolved type parameters.

So if you have a referenced 10s other nuget packages and all have 10s of inline functions, then... all of those might be potentials for your auto-complete. You probably don't want call those, so it's just noise that could be avoided if they'd be internal functions.

Thorium:
Here is quite detailed reference post of Structs: https://www.bartoszsypytkowski.com/writing-high-performance-f-code/

PR #917, opened 2024-05-24
Details improvements

brianrourkeboll:
Fixes (part of) #916.

- Auto-expand the details of the selected member, including when linked directly to (via the fragment portion of the URL).
  
  https://github.com/fsprojects/FSharp.Formatting/assets/14795984/3c8a404a-3754-4f7f-b461-c5a97f3ffc7e

- Keep member name aligned to the top left, i.e., don't shift around when details are expanded. This was the original behavior in #818.

  https://github.com/fsprojects/FSharp.Formatting/assets/14795984/2d527fbf-baba-4153-bf1c-6d71ea16c77b

- Put the details marker back before the summary. (This is how it was before and how `details` normally works... The arrow on its own line as it is in `main` looks very odd to me.)

  image

- Make the table readable on mobile by wrapping rows when the screen is too narrow.

  https://github.com/fsprojects/FSharp.Formatting/assets/14795984/179fc73c-a9e5-468f-93d4-066b604cec95

- Align copy icons, etc., to flex-start instead of [center](https://tonsky.me/blog/centering/) to match member name alignment

  #### Before
  image

  #### After
  image

Issue #916, opened 2024-05-23
Suggestion: auto-expansion of `<details>…</details>` for direct links


brianrourkeboll:
While #778, as implemented in #818, does indeed make for easier skimming of API docs, I think it would be nice if:

- [x] There were some kind of "expand/collapse all" button.

   Yes, I can click expand for every member individually if I want to browse examples, or I can throw

   ```js
   for (const deets of document.getElementsByTagName('details')) {
       deets.setAttribute('open', 'true');
   }
   ```

    in the console, but it would be nice to have a button.

- [x] The `<details>…</details>` for a given member were auto-expanded when linked to directly.

   For example, linking directly to `List.tryPick` like https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html#tryPick currently shows this:

   ![image](https://github.com/fsprojects/FSharp.Formatting/assets/14795984/144a90fe-46c2-4bdb-a9cf-34cf4e098622)

   but I think it would make sense to auto-expand the details for the linked member:

   ![image](https://github.com/fsprojects/FSharp.Formatting/assets/14795984/1a40765c-2c93-4f35-a634-1a34d958355d)

(I'd be willing to open a PR for these.)

nojaf:
Sounds reasonable, go for it!

nojaf:
Hmm, still some remaining problems with the new approach:

https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpchecker.html#InvalidateAll

![image](https://github.com/fsprojects/FSharp.Formatting/assets/2621499/ca3a8eb2-4b6f-4c75-be9b-fe75ff97923d)

brianrourkeboll:
@nojaf Hmm, I'll look.

Separately, I'm noticing else on that same page that will necessitate some additional tweaks in that area anyway.

The `.fsdocs-entity-xmldoc { > div { flex-direction: row reverse } }`, which was already there, affects the order of any paragraphs defined with `<para>` that are nested in the `<summary>`.

So these [paragraphs](https://github.com/dotnet/fsharp/blob/2a65095c471dc557a5720d5c498d9ac18336649e/src/Compiler/Service/service.fsi#L178-L188) —

```fsharp
/// <summary>
/// <para>
///   Parse and check a source code file, returning a handle to the results
/// </para>
/// <para>
///    Note: all files except the one being checked are read from the FileSystem API
/// </para>
/// <para>
[...]

PR #919, opened 2024-06-03
More details/member docs tweaks

brianrourkeboll:
Followup to #917.

- Fix display of member docs for members without other details (e.g., [nullary void-returning methods](https://github.com/dotnet/fsharp/blob/bcffdf70f5d5df156baee6c8e3e8a589bdadd739/src/Compiler/Service/service.fsi#L432-L433)). (https://github.com/fsprojects/FSharp.Formatting/issues/916#issuecomment-2141429522)
- Nest icon buttons in member doc summary in their own `<div>` so that we don't need to use `row-reverse`. This means that any paragraphs (via `<para>`) in the `<summary>` are actually shown in the right order. Paragraphs are also now shown vertically separated instead of horizontally separated. (https://github.com/fsprojects/FSharp.Formatting/issues/916#issuecomment-2141792974)
- Show a source link even when there is no summary XML doc. (https://github.com/fsprojects/FSharp.Formatting/issues/916#issuecomment-2141884878)

### DIFF
--- a/docs/content/fsdocs-default.css
+++ b/docs/content/fsdocs-default.css
@@ -1046,14 +1046,9 @@ span[onmouseout] {
     .fsdocs-entity-xmldoc {
         > div {
             display: flex;
-            flex-direction: row-reverse;
-            justify-content: flex-start;
+            flex-direction: row;
+            justify-content: space-between;
             align-items: flex-start;
-
-            & p.fsdocs-summary {
-                margin: 0;
-                flex-grow: 1;
-            }
 
             & pre {
                 margin-bottom: var(--spacing-200);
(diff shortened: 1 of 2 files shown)


nojaf (approved):
Thanks, this looks good now:

image

@nhirschey, @kMutagene I would like to ship this one as another revision.
Though it technically is a breaking change. The generated HTML for the API documentation is different, so folks which have custom styling for these pages might end up having problems after updating.
Although I'm not really aware of anyone that would face this problem. So, I'm not overly inclined to bump the major version for this. Any objections?

nhirschey:
Shipping as revision sounds good to me. The new styling is still pretty new.

nojaf:
Alright, going to merge this one in and ship a new release with the next version.

PR #920, opened 2024-06-04
Add expand/collapse-all button for API doc details

brianrourkeboll:
Addresses the other suggestion in #916.

- Add button for expanding/collapsing all member doc details.

https://github.com/fsprojects/FSharp.Formatting/assets/14795984/70041395-4893-474d-a297-7651265624c9

I have the buttons operating on the entire document. When toggled, the preference (all expanded or all collapsed) is saved to local storage. An alternative could have been to have independent expand/collapse-all buttons for each section (e.g., constructors instance members, static members, etc.), and no persistence to local storage.

To maintain parity with the behavior of the theme toggle button, I have the new button showing the state that clicking it will _cause_ rather than the current state, even though I think I might personally prefer it the other way around.

### DIFF
--- a/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
@@ -109,7 +109,9 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
                   thead [] [
                       tr [] [
                           td [ Class "fsdocs-member-list-header" ] [ !!tableHeader ]
-                          td [ Class "fsdocs-member-list-header" ] [ !! "Description" ]
+                          td [ Class "fsdocs-member-list-header" ] [
+                              div [ Class "fsdocs-description-heading" ] [ !! "Description"; fsdocsDetailsToggle [] ]
+                          ]
                       ]
                   ]
                   tbody [] [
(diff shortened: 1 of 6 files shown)


src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
```diff
@@ -109,7 +109,9 @@ type HtmlRender(model: ApiDocModel, ?menuTemplateFolder: string) =
                   thead [] [
                       tr [] [
                           td [ Class "fsdocs-member-list-header" ] [ !!tableHeader ]
-                          td [ Class "fsdocs-member-list-header" ] [ !! "Description" ]
+                          td [ Class "fsdocs-member-list-header" ] [
+                              div [ Class "fsdocs-description-heading" ] [ !! "Description"; fsdocsDetailsToggle [] ]
+                          ]
```
brianrourkeboll, line 114: I guess the styling could probably just be applied to `.fsdocs-member-list-header` (which is currently unused), without needing to wrap these in a new `<div>`. Would that be better?

nojaf (approved):
Another solid PR Brian! Thank you!

src/FSharp.Formatting.ApiDocs/GenerateHtml.fs
brianrourkeboll: It seems to work fine: d038a74.

The updated styling shouldn't change the display of any of the other heading cells with this class, since they each only have one child element: https://github.com/search?q=repo%3Afsprojects%2FFSharp.Formatting%20%22fsdocs-member-list-header%22&type=code

PR #922, opened 2024-06-13
Minor Seq Optimizations

1eyewonder:
I was poking around the repo and saw these few spots where I thought we could get some minor performance gains. There are also quite a few areas where we have multiple `Seq.Filter`'s next to each other. I would also like to suggest updating those unless maintainers know of a reason why we have multiple filter iterations next to each other. Hope this helps!

### DIFF
--- a/src/Common/StringParsing.fs
+++ b/src/Common/StringParsing.fs
@@ -109,7 +109,7 @@ module String =
         let spaces =
             lines
             |> Seq.choose (fun line ->
-                if not <| String.IsNullOrWhiteSpace line then
+                if String.IsNullOrWhiteSpace line |> not then
                     line |> Seq.takeWhile Char.IsWhiteSpace |> Seq.length |> Some
                 else
                     None)
--- a/src/Common/StringParsing.fs
+++ b/src/Common/StringParsing.fs
@@ -108,8 +108,11 @@ module String =
     let removeSpaces (lines: string list) =
         let spaces =
             lines
-            |> Seq.filter (String.IsNullOrWhiteSpace >> not)
-            |> Seq.map (fun line -> line |> Seq.takeWhile Char.IsWhiteSpace |> Seq.length)
+            |> Seq.choose (fun line ->
+                if not <| String.IsNullOrWhiteSpace line then
+                    line |> Seq.takeWhile Char.IsWhiteSpace |> Seq.length |> Some
+                else
+                    None)
             |> fun xs -> if Seq.isEmpty xs then 0 else Seq.min xs
 
         lines
--- a/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
@@ -1282,8 +1282,9 @@ module internal TypeFormatter =
                 n
 
         curriedArgs
-        |> List.map (List.map (fun x -> formatArgNameAndType (counter ()) x |> fst))
-        |> List.map (fun argTuple ->
+        |> List.map (fun args ->
+            let argTuple = args |> List.map (formatArgNameAndType (counter ()) >> fst)
+
             match argTuple with
             | [] -> !! "()"
             | [ argName ] when argName = "()" -> !! "()"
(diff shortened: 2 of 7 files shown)


src/Common/StringParsing.fs
```diff
@@ -108,8 +108,11 @@ module String =
     let removeSpaces (lines: string list) =
         let spaces =
             lines
-            |> Seq.filter (String.IsNullOrWhiteSpace >> not)
-            |> Seq.map (fun line -> line |> Seq.takeWhile Char.IsWhiteSpace |> Seq.length)
+            |> Seq.choose (fun line ->
+                if not <| String.IsNullOrWhiteSpace line then
```
nojaf, line 112: Please don't use `<|`

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
```diff
@@ -2264,8 +2264,11 @@ module internal SymbolReader =
         do
             replacedParagraphs
             |> Seq.collect collectParagraphIndirectLinks
-            |> Seq.filter (linkNotDefined doc)
-            |> Seq.map (getTypeLink ctx)
+            |> Seq.choose (fun line ->
+                if linkNotDefined doc line then
```
nojaf, line 2268: Please invert the if/else here, I prefer the negative branch to be the if branch.

src/FSharp.Formatting.Common/Templating.fs
```diff
@@ -48,10 +48,15 @@ type FrontMatterFile =
                 // Allow empty lines in frontmatter
                 let isBlankLine = String.IsNullOrWhiteSpace line
                 isBlankLine || line.Contains(":"))
-            |> Seq.filter (String.IsNullOrWhiteSpace >> not)
-            |> Seq.map (fun line ->
-                let parts = line.Split(":")
-                parts.[0].ToLowerInvariant(), parts.[1])
+            |> Seq.choose (fun line ->
+                let parts = line.Split(":") |> Array.toList
```
nojaf, line 52: This doesn't seem safe when `line` is null.

src/FSharp.Formatting.Common/Templating.fs
```diff
@@ -48,10 +48,15 @@ type FrontMatterFile =
...
+
+                if not <| String.IsNullOrWhiteSpace line then
```
nojaf, line 54: Avoid `<|` please

nojaf (commented):
While I appreciate this PR, the proper thing to do here is most likely a bump of the analyzer tool and packages.
What you detected manually should be caught be a tool.

1eyewonder:
Thanks for the quick review! I will go ahead and make these changes when I get back from travel early next week

nojaf:
Thank you, safe travels!

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
1eyewonder: I updated to use a local variable to prevent a double negative. Let me know if you prefer to not use this

src/FSharp.Formatting.Common/Templating.fs
1eyewonder: Good catch. I moved this within the if check

1eyewonder:
Here are the few areas I see consecutive `Seq.filter`'s used. Do you see any reason why we couldn't consolidate to help reduce iterations? The collections may be short enough where they were broken out for readability but I wanted to check to see if we wanted to include in this PR.

1. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/GenerateModel.fs#L2424-L2425
2. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/GenerateModel.fs#L2451-L2452
3. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/GenerateModel.fs#L2464-L2466
4. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/GenerateModel.fs#L2621-L2630
5. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.Common/YaafFSharpScripting.fs#L183-L184
6. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.Common/YaafFSharpScripting.fs#L284-L294
7. https://github.com/fsprojects/FSharp.Formatting/blob/main/src/FSharp.Formatting.ApiDocs/Categorise.fs#L61-L81

nojaf:
> Here are the few areas I see consecutive Seq.filter/List.filter's used. Do you see any reason why we couldn't consolidate to help reduce iterations? 

This codebase is rather old and has many authors. I don't believe there is a good reason these cases are not combined. Could you address these as well?

src/FSharp.Formatting.ApiDocs/GenerateModel.fs
nojaf: I feel like updating the definition of `linkNotDefined` to `linkDefined` might also be an option. It is only used once and could probably ne simplified to a `List.exist` instead of the map/reduce combo.

nojaf (approved):
Thanks!

Issue #924, opened 2024-06-25
Enable `watch` server from arbitrary host


tymokvo:
While working on a project with a new developer in a GitHub Codespace today, I wanted to be able to serve the documentation of our feature branch. However, I ran into a limitation of the `watch` command where `dotnet tool run fsdocs watch` is incompatible with a remote host because the relative links in the generated HTML are all rooted at `http://localhost:8901`.

Would it be possible to configure `watch` to:
- use an empty string for the `{{root}}` value in HTML templates
- serve the `/tmp/watch` directory
- redirect requests to paths other than `[content | reference]/*` to `index.html`

I think something as simple as `dotnet tool run fsdocs watch --remote` would be very convenient. Though, maybe there is already a way to accomplish this with a combination of existing flags?

nojaf:
Ah yes, this rings a bell. Maybe adding `--parameters root ""` will allow for an empty root and makes things relative but I haven't tested this.

tymokvo:
Hm, that's a start. But it seems that flag works for `build` but not for `watch`. All of the static files generated in `tmp/watch` are still using absolute URLs to `localhost:8901`.

tymokvo:
I think I figured out a patch in #925 . But I'm a little suspicious since there was a whole branch dedicated to preventing a user override of `root` in `watch.

nojaf:
As mentioned in https://github.com/fsprojects/FSharp.Formatting/pull/925#discussion_r1671693970, I would short term try the `--parameters root "/"` in watch mode.

Ideal situation would probably be have `root` be `/` by default (in future versions) and have a `--root` flag to easily set this parameter.

I don't know why root has to be localhost for watch mode. I propose to change this and deal with the real problem once it surfaces. We can also play with this during some alphas.

PR #925, opened 2024-06-26
Remove root parameter override in watch mode

tymokvo:
## Summary
- Remove `if watch` branch when setting `userRoot`

## Description
Re: #924 , it seems this behavior is specifically intended. I'm sure there are other associated changes that this would need. Would love pointers on where to go from here.

### DIFF
--- a/src/fsdocs-tool/BuildCommand.fs
+++ b/src/fsdocs-tool/BuildCommand.fs
@@ -1405,24 +1405,12 @@ type CoreBuildOptions(watch) =
 
         // Adjust the user substitutions for 'watch' mode root
         let userRoot, userParameters =
-            if watch then
-                let userRoot = sprintf "http://localhost:%d/" this.port_option
+            let r =
+                match userParametersDict.TryGetValue(ParamKeys.root) with
+                | true, v -> Some v
+                | _ -> None
 
-                if userParametersDict.ContainsKey(ParamKeys.root) then
-                    printfn "ignoring user-specified root since in watch mode, root = %s" userRoot
-
-                let userParameters =
-                    [ ParamKeys.root, userRoot ]
-                    @ (userParameters |> List.filter (fun (a, _) -> a <> ParamKeys.root))
-
-                Some userRoot, userParameters
-            else
-                let r =
-                    match userParametersDict.TryGetValue(ParamKeys.root) with
-                    | true, v -> Some v
-                    | _ -> None
-
-                r, userParameters
+            r, userParameters
 
         let userCollectionName =
             match (dict userParameters).TryGetValue(ParamKeys.``fsdocs-collection-name``) with


build.fsx
```diff
@@ -93,4 +107,15 @@ pipeline "Verify" {
     runIfOnlySpecified true
 }
 
+pipeline "InstallTool" {
```
nojaf, line 110: I don't quite get why you need this. What exactly are you trying to achieve here?

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -1403,26 +1403,10 @@ type CoreBuildOptions(watch) =
...
-
-                Some userRoot, userParameters
-            else
-                let r =
-                    match userParametersDict.TryGetValue(ParamKeys.root) with
-                    | true, v -> Some v
-                    | _ -> None
-
-                r, userParameters
+        let userRoot =
```
nojaf, line 1406: This will clash later on with

https://github.com/fsprojects/FSharp.Formatting/blob/2efb355823b8311f91a06566c0a3de86037a590d/src/fsdocs-tool/ProjectCracker.fs#L611-L614

Just following the instructions in the readme (`dotnet build` and `src\fsdocs-tool\bin\Debug\net6.0\fsdocs.exe watch`) will launch everything pointing to `https://fsprojects.github.io/FSharp.Formatting` and not localhost.

nojaf (changes requested):
I believe these changes are too disruptive for any existing scenarios.
I think most people don't pass `root` as parameter and that should still work as everybody expects it.

build.fsx
tymokvo: It's not strictly needed. But, my goal was to make it easy to use the same pipeline that is used to test the tool that results from `dotnet pack` so that testing locally and what happens in CI are aligned.

I can revert the `build.fsx` changes if invoking the executable is the recommended way to test locally though.

nojaf: Yes, please revert this. Running the binary locally to test should suffice.

tymokvo: Done!

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -2115,3 +2119,13 @@ type WatchCommand() =
 
     [<Option("port", Required = false, Default = 8901, HelpText = "Port to serve content for http://localhost serving.")>]
     member val port = 8901 with get, set
+
+    [<Option("contenthost", Required = false, HelpText = "URI root to inject in static content.")>]
+    member val contenthost = "" with get, set
+
+    override x.static_content_host_option =
```
nojaf, line 2126: This code violates some analyzers:

image

Run `dotnet fsi build.fsx -- -p Verify` for more info.

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -2115,3 +2119,13 @@ type WatchCommand() =
 
     [<Option("port", Required = false, Default = 8901, HelpText = "Port to serve content for http://localhost serving.")>]
     member val port = 8901 with get, set
+
+    [<Option("contenthost", Required = false, HelpText = "URI root to inject in static content.")>]
+    member val contenthost = "" with get, set
```
nojaf, line 2124: Please add documentation for this new setting.
I'm not quite convinced about the flag name and help text.
I also have some slight concerns that invalid input won't be displayed nicely to the end-users.
If you value for this is `blah`, Suave might trip over it.
It is also not quite clear if `http://` or `https://` needs to be included in this.

nojaf (changes requested):
I think this is heading in the right direction but I do have some remarks.

src/fsdocs-tool/BuildCommand.fs
tymokvo: When you say:

> Please add documentation for this new setting.

Do you mean just in the `HelpText`? Or is there somewhere else that I should be adding a more verbose description?

And good point about the invalid input. I added a call to `System.Uri` in the `override` to get its built-in parse errors. Do you think that is helpful enough?

src/fsdocs-tool/BuildCommand.fs
tymokvo: Thanks for the tip with `-p Verify`! That's helpful. 

Though the output is quite verbose with results from the other projects. It also seems to create diffs in the `tests/` directory for me. Not sure if that's intentional.

src/fsdocs-tool/BuildCommand.fs
nojaf: The HelpText is a great start and I would like you to also add an entry in https://fsprojects.github.io/FSharp.Formatting/commandline.html (see https://github.com/fsprojects/FSharp.Formatting/blob/main/docs/commandline.md).

nojaf: image

Could you take a look if we can show something friendlier than this?

tymokvo: Ah got it! And yes I will look into friendly-ifying the URI error.

tymokvo:
Yes, I noticed that those parameters could become out of sync. But isn't it possible that both `--port` and `--contenturlroot` may be desired?

E.g.
- dev A is pairing with dev B on a branch
- They want to see how their changes affect the docs
- dev A starts the `watch` server bound to `localhost:6001` (overriding the default of `8901`)
- dev A opens an `ngrok` tunnel from `dev-a.ngrok.io` to `localhost:6001`
- dev B navigates to `dev-a.ngrok.io` and hits the suave server

In this case, the static content should all replace `root` with `http://dev-a.ngrok.io/` and the suave server should still bind on `6001`, correct?

tymokvo:
Yes, that would also work, I think. 

I am not very familiar with Suave, but I have used the rust and python abstractions for "just start an HTTP file server in this directory" before. I assume Suave has something similar or it's easy enough to create.

nojaf:
I don't think this would be very Suave related. The bigger issues is that `{{root}}` is used all over the place and would need to transformed to the proper relative url.

nhirschey:
> Thoughts kMutagene nhirschey?

It seems complicated to document or explain contentroot, root, and port simultaneously. It's a bit of an edge case that anybody would use these together, and maybe a suboptimal edge case is ok, but still a bit of a hack.

If I understand the problem correctly, a single --root and then relative paths seems the cleanest way forward if it's possible. More work to get it right, but perhaps a better long-term foundation. I don't know why it's absolute rather than relative paths now.

kMutagene:
Agreed with everything that has been said so far, but moving PRs to the 'More work to get it right, but perhaps a better long-term foundation' solution tends to stall PRs such as this - see my still-open PR of shame on this repo.

@tymokvo would you be up for escalating the scope of this PR that much? Otherwise i'd suggest publishing a 'good enough' solution only in preview versions, and another maintainer/contributor picking up from there in future PRs.

tymokvo:
I would be happy to contribute a better long-term solution! But I was hoping to be able to get this merge-ready before the deadline I have at work now. So, I wouldn't be able to resume this for a couple weeks.

As long as you all are ok with this PR going stale for that amount of time, then I can revisit.

And, just so that I'm clear, the direction would be:
- Support a single `root` flag to declare the URL root
- Remove the existing override in the `watch --root` case
- Update the path generation in web content to all use relative paths over absolute paths to `{{root}}`

...correct?


My main concern would be the risk of breaking changes since the scope would be so broad and would affect any project that depends on `FSharp.Formatting`.

nojaf:
Hi @tymokvo, thanks for the offer. We have a call with the maintainers tomorrow, I'll bring it up over there.

src/fsdocs-tool/BuildCommand.fs
```diff
@@ -1406,10 +1404,13 @@ type CoreBuildOptions(watch) =
         // Adjust the user substitutions for 'watch' mode root
         let userRoot, userParameters =
             if watch then
-                let userRoot = sprintf "http://localhost:%d/" this.port_option
+                let userRoot =
```
nojaf, line 1407: Would

```fsharp
let userRoot =
    match userParametersDict.TryGetValue(ParamKeys.root) with
    | true, root -> root
    | false, _ -> sprintf "http://localhost:%d/" this.port_option
```

work for you here?
I believe it can be an easy way to make 
```
fsdocs watch --parameters root "/"                                                                                                                                                                               
```

work and would be the least disruptive for now.

We don't feel good enough about the new setting.

tymokvo: I will test and let you know! Thanks for talking it through!

nojaf: Sure thing, thanks for your patience!

tymokvo:
Would love some help. Unfortunately work is consuming all my time right now.

dsyme:
@tymokvo I think we can close this out right new - recreate an issue with a really clean spec of the change we want and then set agentic analysis investigating it. 

Good luck with the work

Issue #929, opened 2024-07-09
Use signature-like for documenting the API


MangelMaxime:
I propose to revisit the documentation to use "signature-like" to describe the API documentation. This is also to makes the most out of the XML doc comment.

Here is a comparaison between what I propose and the currently generated documentation by F# formatting using Fable.Promise package as a reference. Fable.Promise has been crafted to take the most out of the XML doc comment.

![CleanShot 2024-07-09 at 21 49 20](https://github.com/fsprojects/FSharp.Formatting/assets/4760796/46396227-fc0c-4e57-a1a2-e726bda3e3dc)

![CleanShot 2024-07-09 at 21 51 28](https://github.com/fsprojects/FSharp.Formatting/assets/4760796/6cf31696-6a02-4a4f-baaa-66cfed4df67a)

> [!NOTE]
> The goal of this issue is to focus on improving the information presented to the user. Don't focus too much on the global layout / navigation differences.

- Instead of using a table like layout with the `function` or `member` name on the left with the documentation on the right, the new proposition use a vertical layout.
- The tooltip formatting can be improved by re-using the formatting used in FsAutoComplete.

Here is another exemple for a record type:

![CleanShot 2024-07-09 at 22 12 50](https://github.com/fsprojects/FSharp.Formatting/assets/4760796/d0ac7579-ed10-4527-9b11-9ab7867a03c3)

To have a look at my proposition, you can go to [Fable.Promise](https://fable.io/fable-promise/reference/Fable.Promise/index.html). It mainly use functions.

If you want to see more complex API documentation with a mix of functions and records, you can clone https://github.com/MangelMaxime/Nacara/tree/fsharp-docs-example and then use `npx http-server` to host it locally. It is not available online anymore, because of a broken thing since the project moved to NET6.0

I will have a look in the coming days, to send a pull request to see what it means in term of changes of F# Formatting.

brianrourkeboll:
I'd definitely be interested in seeing a prototype of this.

> * Instead of using a table like layout with the `function` or `member` name on the left with the documentation on the right, the new proposition use a vertical layout.

<details>
<summary>
The current layout does technically switch to vertical when the viewport is below a certain width 😉
</summary>

![image](https://github.com/fsprojects/FSharp.Formatting/assets/14795984/5feacdda-80e1-4ee9-a566-383fe3ff50ec)

</details>

<details>
<summary>
[...]

PR #932, opened 2024-07-13
[WIP] Experiment with porting generation to signature like API

MangelMaxime:
**THIS IS NOT READY FOR REVIEW**

Right now, I am just hacking my way to port my code and check what difficulties can be encountered with implementing #929.

The current state of this PR, doesn't reflect the final proposition as I am just adding everything needed in place. Later on, I will refactor the code to make it more readable and maintainable.


nhirschey:
@MangelMaxime  just an FYI in case it's useful. Not trying to direct you one way or another, there was a prior attempt at signatures that got abandoned when branches got out of sync: https://github.com/fsprojects/FSharp.Formatting/pull/547 

Again, just FYI in case there's something useful there.

MangelMaxime:
@nhirschey Thanks for pointing out.

Looks like the problem reported with F# Formatting introducing whitespace where we don't want was there too.

The main difference I see beside the code used to render the signature, is that they focus on adding it in the tooltip while in my proposition we don't need tooltip anymore. This also allows to have part of the signature easily clickable for example,  you can click `Promise` from `Promise<'T>` to be redirected to `Fable.Core.JS.Promise` documentation (the URL needs to be fixed or the documentation of dependencies generated too).

nhirschey:
> while in my proposition we don't need tooltip anymore. 

I might be misunderstanding what you mean, but even if we don't *need* the signature tooltip in ApiDocs anymore we'd still use the signature tooltip in literate docs files like [these](https://fsprojects.github.io/FSharp.Formatting/literate.html). I'd probably say keep the signature tooltip in both literate docs and apidocs.

> Looks like the problem reported with F# Formatting introducing whitespace

Yeah, the whitespace is a pain to keep track of in various tests. Agreed it could be nicer. You may see in the tests various adhoc functions to strip white space.

We're targetting the commonmark [spec](https://[spec.commonmark.org/0.31.2/spec.json](https://spec.commonmark.org/0.31.2/spec.json)) when converting markdown to html. I'm not sure how minified html affects that, but worth keeping in mind.

MangelMaxime:
> I might be misunderstanding what you mean, but even if we don't _need_ the signature tooltip in ApiDocs anymore we'd still use the signature tooltip in the code blocks of literate docs files like [these](https://fsprojects.github.io/FSharp.Formatting/literate.html). I'd probably say keep the signature tooltip in both literate docs and apidocs.

Signature tooltips are kept in the literate docs.

For the API docs, we don't need to keep them because all the information is already in the new signature and documentation.

One thing to decide, could be if we want to always show all the information like Elm:

CleanShot 2024-07-15 at 15 26 40

Show a portion of it and add a `ReadMore` button like Rust:

CleanShot 2024-07-15 at 15 27 14

Make it fully collapsed like VSCode API:

CleanShot 2024-07-15 at 15 27 51


CleanShot 2024-07-15 at 15 18 14

CleanShot 2024-07-15 at 15 19 10

> We're targetting the commonmark [spec](https://%5Bspec.commonmark.org/0.31.2/spec.json%5D(https://spec.commonmark.org/0.31.2/spec.json)) when converting markdown to html. I'm not sure how minified html affects that, but worth keeping in mind.

I believe that Markdown is not be affected by having a minified or indented HTML.

nojaf:
Hi there, thanks for this initial attempt! Here are some preliminary remarks; I haven't looked into the details yet.

> Improve XML Doc formatting

I'd like to hear the strategy for how Ionide and FSDocs can use the exact same code base. Simply updating the code and keeping a copy doesn't seem like a good idea. Please go the extra mile on this.

> Rendering of Signature

The way the signature is formatted does look quite different from anything I've seen before. I'm not against it, but I'd feel more comfortable if others could weigh in and confirm that this is truly an improvement. It's a departure from the style guide, which isn't necessarily bad, but it should be a very conscious decision. @dsyme, any thoughts on this?

> Remarks regarding the current CSS

he CSS approach was inspired by https://www.refactoringui.com/
The book goes into detail why `px` is used over `em/rem`.
And some of the other choices that were made were inspired by this book.

<summary>
<details>
In the book "Refactoring UI," the authors recommend using px over rem or em for several reasons:

Consistency: px units provide a consistent size across different elements and contexts. This consistency makes it easier to design and maintain a cohesive look and feel throughout the application.

Predictability: px units are more predictable because they do not depend on the font size of parent elements or the root element. This predictability simplifies the design process and reduces unexpected changes in element sizes due to changes in font sizes.

Ease of Design: Designers typically think in terms of pixels when creating mockups and layouts. Using px units allows for a more straightforward translation from design tools to code, ensuring that the implemented design matches the original intention.

Simplified Calculations: Using px units can simplify the calculations needed for spacing, alignment, and sizing. With rem or em, you often have to consider the inheritance and cascading of font sizes, which can complicate these calculations.

Avoiding Accessibility Issues: While rem and em units are often recommended for their scalability and accessibility benefits (like respecting user font size preferences), the book suggests that modern browsers and devices handle pixel-based layouts well enough that using px does not inherently cause accessibility issues, especially if other best practices for accessibility are followed.

The recommendation is based on the practical aspects of designing and maintaining a UI, emphasizing simplicity and reliability in the design-to-code workflow.
</details>
</summary>

There was a specific period for giving feedback on the CSS of the modern theme, from November to February. For me, that window has closed, and I prefer not to revisit it just because a new contributor might have done things differently. This is a common scenario in open source projects, and most of your comments seem to reflect a different perspective rather than actual issues. Discussing this with every new contributor isn't productive, so I don't want to reopen these topics as part of this PR. However, you should be able to contribute your ideas with only localized changes.

> How was the colour for the syntax highlighting chosen?

A bit random, I would be ok with revisiting that.

> Fantomas configuration

I'm not against revisiting the Fantomas configuration. Feel free to propose some changes in a separate PR, and the others can weigh in on whether it's worth making those changes.

> Feliz like syntax for HTML

I'm personally a fan of the two lists DSL; it mimics the XML/JSX structure with attributes and children nicely. This isn't a problem that needs addressing, just a personal preference. There's no need to update the DSL for your changes, so please make it work with what we have.

> Tests

Some regression tests at the end would be nice. It would be great if they could cover everything possible with XML comments. I often struggle to reproduce certain issue reports because I'm less aware of where things are coming from.

MangelMaxime:
> > Remarks regarding the current CSS
> 
> he CSS approach was inspired by [refactoringui.com](https://www.refactoringui.com/) The book goes into detail why `px` is used over `em/rem`. And some of the other choices that were made were inspired by this book.

Thanks for the information, I will have a look at it because this the first time I see `px` being recommended over `rem`.

And like you said, there was a window for CSS feedback that was missed. In theory, because F# formatting allows custom templates hopefully it means that if someone wants it is possible to change everything to their liking. It needs to be confirmed because by a quick glance at the documentation it seems more like it gives HTML output instead of raw string informations but I didn't play with this feature enough yet to be sure of what I am saying.

> > Feliz like syntax for HTML
> 
> I'm personally a fan of the two lists DSL; it mimics the XML/JSX structure with attributes and children nicely. This isn't a problem that needs addressing, just a personal preference. There's no need to update the DSL for your changes, so please make it work with what we have.

I have the same idea on this point indeed. And improving the Fantomas configuration will already be a huge saver.

> > Tests
> 
> Some regression tests at the end would be nice. It would be great if they could cover everything possible with XML comments. I often struggle to reproduce certain issue reports because I'm less aware of where things are coming from.

I do have a bunch of tests available for the XML comment formatter in Nacara which I will be able to port over in this PR or in the re-usable library for both FSAC and F# Formatting (or any F# project actually).

> > Rendering of Signature
> 
> The way the signature is formatted does look quite different from anything I've seen before. I'm not against it, but I'd feel more comfortable if others could weigh in and confirm that this is truly an improvement. It's a departure from the style guide, which isn't necessarily bad, but it should be a very conscious decision. @dsyme, any thoughts on this?

Something to note here is based on the snippet provided by @nhirschey 

> ```fsharp
> val eitherEnd:
>     success: 'T -> unit ->
>     fail: exn -> unit ->
>     pr: 'T ->
>         unit
> ```

Then generating this kind of signature is easier because we don't need to know what is the max width of the different elements to keep the alignement.

I personally prefer the aligned version of the signature for documentation because it easier on the eyes thanks to blocking. And also because it is done automatically, I would not do that inside of my code as I don't want to spend time manually indenting stuff 😅

Now that I gathered several feedbacks and I am back from vacation, I will continue working on this PR.

MangelMaxime:
Coming back regarding the Signature format, I just remembered that I took inspiration on how Ionide renders it:

CleanShot 2024-08-13 at 15 26 11

.editorconfig
```diff
@@ -16,5 +16,21 @@ fsharp_max_array_or_list_width=120
 fsharp_max_infix_operator_expression=80
 fsharp_max_value_binding_width=120
 
-[GenerateHtml.fs]
-fsharp_experimental_elmish = true
\ No newline at end of file
+[{GenerateHtml.fs,tests/FSharp.ApiDocs.Tests/files/ReferenceProject/**/*.fs}]
```
nojaf: I'd prefer a separate PR for this.

build.fsx
```diff
@@ -93,4 +93,38 @@ pipeline "Verify" {
     runIfOnlySpecified true
 }
 
+let referenceProjectDir = "./tests/FSharp.ApiDocs.Tests/files/ReferenceProject"
+
+// TODO: Revisit to see how we can use `dotnet watch` to run a local version of the tool
```
nojaf, line 98: This worked for me:

```bash
dotnet watch run --project .\src\fsdocs-tool\fsdocs-tool.fsproj -- watch --input "$HOME/projects/fantomas/docs" --eval --projects "$HOME/projects/fantomas/src/Fantomas.Core/Fantomas.Core.fsproj"
```

(you probably don't need `--eval`)

MangelMaxime: Thanks for the hint I will have a look.

.editorconfig
MangelMaxime: Sure will do, just wanted to have it at least for me when working locally because I was too frustrated with dealing with indentation manually 😅

brianrourkeboll:
https://github.com/fsprojects/FSharp.Formatting/pull/932#issuecomment-2273050820

> > rendering of signature.
> 
> I don't have strong feelings but I'd probably stick with the style guide. My interpretation of the style guide is the rendering should be
> 
> ```fsharp
> val eitherEnd:
>     success: 'T -> unit ->
>     fail: exn -> unit ->
>     pr: 'T ->
>         unit
> ```

https://github.com/fsprojects/FSharp.Formatting/pull/932#issuecomment-2284007628

> I personally prefer the aligned version of the signature for documentation because it easier on the eyes thanks to blocking. And also because it is done automatically, I would not do that inside of my code as I don't want to spend time manually indenting stuff 😅

https://github.com/fsprojects/FSharp.Formatting/pull/932#issuecomment-2286254241

> Coming back regarding the Signature format, I just remembered that I took inspiration on how Ionide renders it:
> 
> CleanShot 2024-08-13 at 15 26 11

For what it's worth, I also subjectively find the aligned signatures easier to read (especially for longer or more complex ones), and I think that they are a good fit for automatically generated documentation like this, even though they differ from the style guide.

I always appreciate Ionide's (aligned) signature tooltips every time I use it instead of VS (not a completely fair comparison, since the style guide does at least allow for indenting params, while VS tooltips just wrap arbitrarily).

The style guide's main motivation for _not_ recommending aligning signatures in source code — minimizing diffs/manual realignment — doesn't apply here, and the added readability for something whose sole purpose is to be read seems (to me) more important than strict consistency with the style guide.

But I can understand not wanting to add a new inconsistency, however small and justified, to a core library like this.

nhirschey:
Good feedback @brianrourkeboll, thank you. I’m fine with alignment, but I’d say stick with style-guide suggestion that all lines end with ->, particularly the last line. 

use 
```fsharp
pr: 'T ->
        unit 
```

Instead of 
```fsharp
pr: 'T
        -> unit 
```

I assume the style guide made that point about `->` for a reason. 

All that said, whatever ionide does carries weight for sure. The signature format of ionide is the way many people are used to seeing them and will get interpreted as the de facto standard.

MangelMaxime:
I finally have some time available to work on this again, here is the new preview based on discussion we had with Florian.

CleanShot 2024-12-30 at 11 36 52@2x

PS: The code from this PR is not representative of the final proposition, I am mostly making my way to see if the result is of interest for this project.

dsyme:
@MangelMaxime I'll close this out since it's been stale a year and has merge conflicts.  Please do reopen if you'd like to resurrect

Issue #799, opened 2023-02-23
Consider targetting .NET 6 as well


kMutagene:
since v17, `fsdocs-tool` only supports .NET 7. This can cause problems especially in projects where FAKE is being used, as FAKE targets .NET6 and will not work due to incompatible msbuild log formats when running with the .NET 7 sdk. Should there be no .NET 7 exclusive code i'd suggest multi-targetting frameworks here:

https://github.com/fsprojects/FSharp.Formatting/blob/e665aa95ab897e5e2c543e5847f3820b3e215e92/src/fsdocs-tool/fsdocs-tool.fsproj#L6

to 

```xml
 <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
```

or even only targetting .net 6, as that is the LTS release anyways

nojaf:
Hello,

Thank you for bringing this up. This may not be possible because we depend on MSBuild bits for project cracking. @baronfel any thoughts?

TheAngryByrd:
I think it's possible but we need to set the `Microsoft.Build` to `copy_local: false`. 

```
# copy_local: false excludes runtime assets (desired for msbuild)
nuget Microsoft.Build copy_local: false
nuget Microsoft.Build.Framework copy_local: false
nuget Microsoft.Build.Utilities.Core copy_local: false
nuget Microsoft.Build.Tasks.Core copy_local: false
```


https://github.com/ionide/proj-info/issues/178

nojaf:
We would first need to have https://github.com/fsprojects/FSharp.Formatting/pull/796 right?
As our current version of FCS would pull down those dependencies anyway.

TheAngryByrd:
> We would first need to have #796 right? As our current version of FCS would pull down those dependencies anyway.

I don't think we need it for just this. The [dependencies for Ionide.ProjInfo.FCS](https://www.nuget.org/packages/Ionide.ProjInfo.FCS#dependencies-body-tab) are `FSharp.Compiler.Service(>= 41.0.5)`

<img width="304" alt="image" src="https://user-images.githubusercontent.com/1490044/221181528-fea65a09-914d-431f-9875-93b4377b5b76.png">

nojaf:
Yeah, until you actually install it.
Tried this console app:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
[...]

PR #939, opened 2024-09-01
Target .NET 8 in the tool.

teo-tsirpanis:


nojaf:
Any particular reason you want this?
We have the [RollForward](https://learn.microsoft.com/en-us/dotnet/core/versions/selection#control-roll-forward-behavior) in place. `<RollForward>LatestMajor</RollForward>` ensures that a .NET tool or application uses the latest installed major version of the runtime, even if it differs from the targeted version.

We rolled it back to 6.0 after https://github.com/fsprojects/FSharp.Formatting/issues/799

teo-tsirpanis:
OK. I had opened this PR almost instinctively, after discovering having made the change in an old working copy of mine without committing it. I did not see the `RollForward`.

PR #943, opened 2024-11-02
note on TargetFramworks ( the plural)

goswinr:
I believe my edit of the readme is correct, See :    https://github.com/fsprojects/FSharp.Formatting/blob/d6e29d397ff44dbcf402b476b63340b217251e12/src/fsdocs-tool/ProjectCracker.fs#L390


nojaf (approved):
Sounds about right, thanks!

PR #948, opened 2024-12-30
style: try make code more consistent and align HTML DSL to 4 spaces indentation

MangelMaxime:
Like requested in #932 this is a PR dedicated to the Fantomas configuration.

When working on a new structure for the API documentation, I discovered that the current setup make it difficult to work with the HTML dsl.

Indeed, the elements inside of the list of the DSL are not naturally aligned to the standard 4 spaces indentations so you often need to manually place the code by adding 1, 2, 3 spaces, etc.

The important rules are to avoid the indentation issue are:

```editorconfig
fsharp_experimental_elmish = true
fsharp_array_or_list_multiline_formatter = number_of_items
fsharp_max_array_or_list_number_of_items = 0
fsharp_multi_line_lambda_closing_newline = true
fsharp_multiline_bracket_style = aligned
```

The rest of the rules are here to enforce more consistency overall but are probably more subject to debate. I just wanted to take this opportunity of updating the Fantomas config to open the discussion.

For this PR, I applied the new rules to `GenerateHtml.fs` only but we could and perhaps should apply them to all the files for consistency ?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,10 +22,5 @@ fsharp_max_array_or_list_number_of_items = 0
 fsharp_multi_line_lambda_closing_newline = true
 fsharp_multiline_bracket_style = aligned
 fsharp_keep_max_number_of_blank_lines = 1
-fsharp_align_function_signature_to_indentation = true
 # Keep the if/then/else consistent
 fsharp_max_if_then_else_short_width = 0
-# Keep records consistent
-fsharp_max_record_number_of_items = 0
-fsharp_record_multiline_formatter = number_of_items
-fsharp_alternative_long_member_definitions = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,10 +3,7 @@ root = true
 
 [*]
 end_of_line = lf
-
-[*.{fs,fsi,fsx}]
 charset = utf-8
-end_of_line = lf
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
 fsharp_max_infix_operator_expression=80
 fsharp_max_value_binding_width=120
 
+# Specific settings for "View/HTML" related files
+# It makes the code more consistent in term of spacing and indentation
 [GenerateHtml.fs]
 fsharp_experimental_elmish = true
+fsharp_array_or_list_multiline_formatter = number_of_items
+fsharp_max_array_or_list_number_of_items = 0
+fsharp_multi_line_lambda_closing_newline = true
+fsharp_multiline_bracket_style = aligned
+fsharp_keep_max_number_of_blank_lines = 1
+fsharp_align_function_signature_to_indentation = true
+# Keep the if/then/else consistent
+fsharp_max_if_then_else_short_width = 0
+# Keep records consistent
+fsharp_max_record_number_of_items = 0
+fsharp_record_multiline_formatter = number_of_items
+fsharp_alternative_long_member_definitions = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,10 @@ root = true
 
 [*]
 end_of_line = lf
+
+[*.{fs,fsi,fsx}]
 charset = utf-8
+end_of_line = lf
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
@@ -17,10 +20,6 @@ fsharp_max_value_binding_width=120
 # It makes the code more consistent in term of spacing and indentation
 [GenerateHtml.fs]
 fsharp_experimental_elmish = true
-fsharp_array_or_list_multiline_formatter = number_of_items
-fsharp_max_array_or_list_number_of_items = 0
 fsharp_multi_line_lambda_closing_newline = true
 fsharp_multiline_bracket_style = aligned
 fsharp_keep_max_number_of_blank_lines = 1
-# Keep the if/then/else consistent
-fsharp_max_if_then_else_short_width = 0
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
...
 fsharp_max_value_binding_width=120
 
+# Specific settings for "View/HTML" related files
+# It makes the code more consistent in term of spacing and indentation
 [GenerateHtml.fs]
-fsharp_experimental_elmish = true
\ No newline at end of file
+fsharp_experimental_elmish = true
+fsharp_array_or_list_multiline_formatter = number_of_items
+fsharp_max_array_or_list_number_of_items = 0
```
nojaf, line 21: I don't really like this one, it makes too many single item list format in three lines. Makes it all rather excessive.

.editorconfig
```diff
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
...
+fsharp_multi_line_lambda_closing_newline = true
+fsharp_multiline_bracket_style = aligned
+fsharp_keep_max_number_of_blank_lines = 1
+fsharp_align_function_signature_to_indentation = true
```
nojaf, line 25: I'm not seeing the effects from this in this file.

.editorconfig
```diff
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
...
+# Keep the if/then/else consistent
+fsharp_max_if_then_else_short_width = 0
```
nojaf, line 27: I don't think this really solves any problem.

.editorconfig
```diff
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
...
+# Keep records consistent
+fsharp_max_record_number_of_items = 0
```
nojaf, line 29: I don't think this really solves any problem.

.editorconfig
```diff
@@ -16,5 +13,19 @@ fsharp_max_array_or_list_width=120
...
+fsharp_record_multiline_formatter = number_of_items
+fsharp_alternative_long_member_definitions = true
```
nojaf, line 31: Not sure if used.

nojaf (changes requested):
I'd like a subset of these changes that focus on the DSL.
Otherwise it is just very inconsistent toward the rest of the codebase.

.editorconfig
MangelMaxime: This solves consistency in formatting:

CleanShot 2025-01-02 at 11 04 30@2x

Personally, I value having a single way of reading the code compared to have sometimes something take 1 lines, sometimes 2 or more.

Can disable if still wanted

.editorconfig
MangelMaxime: `fsharp_multiline_bracket_style = aligned` is need to ensure the code inside of the DSL is aligned to the 4 spaces guides

CleanShot 2025-01-02 at 11 09 38@2x

`fsharp_keep_max_number_of_blank_lines = 1` has effet on code like that:

CleanShot 2025-01-02 at 11 08 12@2x


`fsharp_align_function_signature_to_indentation = true` does nothing indeed, probably a left over copy/paste from experimentation/my own configuration. Will remove it

.editorconfig
MangelMaxime: This makes the code more consistent it always align to the 4 spaces guides and avoid having to reformat the list over and over when adding a new element:

CleanShot 2025-01-02 at 11 10 52@2x

For example, if I want to add a new element to `embed` with that configuration I can simply press `Enter` and be ready with it. Without I, need to move the list, align it, split the inner value on multiple lines, etc. Or add `;` to add the new item but if it is too long I need to go back manually placing stuff around.

.editorconfig
MangelMaxime: Indeed, it does nothing in this file because it doesn't have any Record.

In the general concept, it is once again for consistency all record are formatted the same way thanks to that configuration. But will remove it.

.editorconfig
MangelMaxime: Does nothing in this file

MangelMaxime:
@nojaf Thanks for the review, I made a first pass to remove the rules which had no effect in this file.

Waiting for feedback on the rest, as I explained why the rule exist or not. Will adapt the PR accordingly :)

.editorconfig
```diff
@@ -3,10 +3,7 @@ root = true
 
 [*]
 end_of_line = lf
-
```
nojaf, line 6: Please don't remove `[*.{fs,fsi,fsx}]`, honestly rule of thumb change as little as possible.

.editorconfig
nojaf: pressing `;` and formatting isn't the end of the world.
How many times are really facing this issue? If you had this once I really don't feel like this is worth it. If you have this 20 times in an hour I could kinda get it, but even then...

.editorconfig
nojaf: Please remove

nojaf (changes requested):

.editorconfig
MangelMaxime: These settings make sense for all the files but ok

.editorconfig
MangelMaxime: More often than I want to have to deal with it 🙂

MangelMaxime: Removing this code also make it harder to read some line IHMO.

CleanShot 2025-01-02 at 20 14 26@2x

I removed it but we can re-introduce if wanted before merging.

MangelMaxime:
@nojaf I think I made all the changes requested.

Last ones that could change perhaps are:

- https://github.com/fsprojects/FSharp.Formatting/pull/948#discussion_r1901185473
- https://github.com/fsprojects/FSharp.Formatting/pull/948#discussion_r1901182270

If no more changes required this PR should be ready.

nojaf (approved):
Thanks!

PR #950, opened 2024-12-30
fix: remove duplicated `Type Extensions`

### LINKED ISSUE #941: One extension members shows four times in docs.
I made an extension member called `.Last`  with `get` and `set` for `'T[]`.
It shows up four times in the generated docs. Can I fix this somehow ?
image
https://goswinr.github.io/ArrayT/reference/arrayt-autoopenarraytextensions.html

MangelMaxime:
Fix #941

It seems like they all have the same `DeclarationLocation` which can make sense if this is just syntax sugar.

### DIFF
--- a/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
+++ b/src/FSharp.Formatting.ApiDocs/GenerateModel.fs
@@ -2737,6 +2737,11 @@ module internal SymbolReader =
             let exts, nsdocs2 =
                 readMembers ctx entityUrl ApiDocMemberKind.TypeExtension modul (fun v -> v.IsExtensionMember)
 
+            // `with get and set` syntax is sugar for a mutable field, a get binding and a set binding
+            // This result in duplicated Method Extensions, we use DeclarationLocation to keep only one
+            // See https://github.com/fsprojects/FSharp.Formatting/issues/941
+            let exts = exts |> List.distinctBy (fun m -> m.Symbol.DeclarationLocation)
+
             let pats, nsdocs3 =
                 readMembers ctx entityUrl ApiDocMemberKind.ActivePattern modul (fun v -> v.IsActivePattern)
 


nojaf:
@goswinr could you check if this PR solves your problem?

MangelMaxime:
To test the change locally, I did something like that:

Adding this code to F# Formatting code, for example at `src/FSharp.Formatting.Literate/Literate.fs`

```fs
module Test =

	type ``[]``<'T>  with //Generic Array
	        member inline xs.First
	            with get () =
					failwith ""
	            and set (v: 'T) =
					failwith ""
```

and generated the docs locally.

Doing so, I was able at the time to reproduce both the issue and the fix.

nojaf:
Would like to Goswin (or anyone really) to try this one the actual project.
There were some changes in the compiler with `get/set` the past years, so would like some confirmation this solves the reported problem.

goswinr:
@MangelMaxime thank you for fixing this. Your test case looks just like mine. Did you test on net9?  @nojaf [I am using](https://github.com/goswinr/ArrayT/blob/5908fbfbf98f0fa68df6c4ac48c2ffbe5b31ff0b/.github/workflows/docs.yml#L27) the latest net9 skd.

MangelMaxime:
@nojaf In theory, my test was a reproduction but I just tested it locally against @goswinr original project:

**Before (online version)**

CleanShot 2025-01-02 at 17 38 36@2x

**After (local with the fix)**

CleanShot 2025-01-02 at 17 51 58@2x

I don't know why the styling is broken in the in the after version, this is perhaps caused by because i run `fsbuild` in build mode and host a standalone http server, so I don't respect the `baseUrl` expected (`http://192.168.0.197:8080/` VS `https://goswinr.github.io/ArrayT`). For example, I see some `.js` file failing to load because of that.

But I didn't touch the CSS in this PR so I didn't spend time trying to fix this "non issue".

nojaf (approved):
Alright, thank you!

PR #953, opened 2025-04-22
Add --ignoreuncategorized flag

simontreanor:
Sometimes it is necessary to link to documents that you do not want to appear on the navigation sidebar, perhaps as they are too numerous and therefore swamp the menu structure.

I've added an `ignoreuncategorized` flag to the build arguments, the effect of which is for the menu to ignore any documents that do not have any category defined in their front matter. It does not prevent the documents' contents from showing, just that they do not appear in the sidebar.

Usage e.g.: `dotnet fsdocs build --clean --properties Configuration=Release --eval --ignoreuncategorized=true`

This flag defaults to `false` if omitted.


nojaf:
Hello, this seems reasonable. However, I would prefer to use `--ignoreuncategorized` without `=true`. Does the absence of the flag set the value to false?  
@nhirschey, what are your thoughts?  

If we add this, I would like to see a reference to it in our [own documentation](https://fsprojects.github.io/FSharp.Formatting). Otherwise, nobody will ever know it exists.

simontreanor:
Hi Florian, thanks for the feedback. I've just checked that the flag can be used on its own, without `=true`, and also that omitting it sets the value to false. I've also added it to the documentation.

simontreanor:
I've taken out the no-wrap span styling, which I think is the simplest remedy.

nojaf:
@simontreanor would you like a release after this is merged?

nojaf:
> That would be great, thank you!

Could you add

```
## 21.0.0-beta-005 - 2025-04-23

### Added
* Add --ignoreuncategorized flag. [#953](https://github.com/fsprojects/FSharp.Formatting/pull/953)
```

to the `RELEASE_NOTES.md` file?

(I couldn't push to your branch because you used your main branch, avoid using your default branch when creating PRs using GitHub in general)

simontreanor:
Done, and duly noted. Thank you for all your help!

nojaf:
New beta available on NuGet: https://www.nuget.org/packages/fsdocs-tool/21.0.0-beta-005
Let me know if this works for you and we can probably call 21 stable.

simontreanor:
Seems to be working fine. I was updating documentation for FSharp.Finance.Personal and included the output of ~250 unit tests as reference ([see here](https://simontreanor.github.io/FSharp.Finance.Personal/unitTestOutput.html)). The navigation menu is a whole lot tidier now without those 250 other entries!

PR #954, opened 2025-04-22
update readme dev instructions to use net8.0

nhirschey:
update readme dev instructions to use net8.0 so that when you copy and paste into terminal it'll use the right binary for the fsdocs tool.


nojaf (approved):
Thanks!

PR #958, opened 2025-09-08
Run analyzers for all projects

nojaf:
GitHub no longer accepts multiple sarif files during upload.
In this PR, I'm trying to run analysis for all projects in one go.


nojaf:
Didn't fix absolutely everything, but this is now again in a good state and should no longer block CI.

nhirschey (approved):
Seems good to me. I built from this PR and looked at the page in watch mode and all seemed fine.

Issue #1054, opened 2026-02-28
Update ProjInfo to v0.74.2 to fix VisualTree.relativePathOf issue


TheAngryByrd:
See: 
https://github.com/ionide/FSharp.Analyzers.SDK/pull/290
https://github.com/ionide/proj-info/pull/244

We should update to: https://github.com/ionide/proj-info/releases/tag/v0.74.2

TheAngryByrd:
/repo-assist investigate and create a PR updating projinfo

TheAngryByrd:
looks like I don't have permission to get it to do things. I need an adult. @dsyme @nojaf 

<img width="1377" height="988" alt="Image" src="https://github.com/user-attachments/assets/70a8803a-8954-4018-b8ae-81b66c6415cb" />

Smaug123:
I've got Claude running for this already - will let you know if it works.

nojaf:
/repo-assist investigate and create a PR updating projinfo

PR #1055, opened 2026-02-28
Upgrade Ionide.ProjInfo to 0.74.2

Smaug123:
PR by Claude (Opus 4.6).

Fixes https://github.com/fsprojects/FSharp.Formatting/issues/1054 .

My knowledge of the MsBuild packages is exceptionally bad; I've tried understanding how it works in the past and simply failed. Claude already got this wrong once in a way CI caught but which I couldn't predict would fail.


nojaf:
@Smaug123 did you try this out somewhere?
Happy to take this one over #1056 , @TheAngryByrd could you take a look at this PR as well please.

If we have this, it would unblock you if we have a new version right?

Smaug123:
I was out this morning, haven't had a chance to try it out plugged into the FSharp.Analyzers.SDK build; might not get a chance to do that for a few hours. I didn't actually include the bugfix I specifically want in this PR; that would be an additional two-line change to convert "find `.sln` files" into "find `.sln` and `.slnx` files", but I thought it would be more hygienic to isolate the two changes.

PR #1140, opened 2026-04-02
Fix spurious 'fsi is not defined' error during literate script type-checking

### LINKED ISSUE #1139: Spurious 'fsi is not defined' error during literate script evaluation
## Description

`fsdocs build --eval` reports a spurious error when a literate script uses `fsi.AddPrinter`:

```
docs/test.fsx: error(8,0)-(8,3) The value, namespace, type or module 'fsi' is not defined.
```

The error is **spurious** — the `fsi.AddPrinter` call **does execute correctly** at runtime and the registered printer **is used** in the output. The error appears to come from the compile-time type-checking pass but not from actual evaluation.

## Minimal repro

Create `docs/test.fsx`:

```fsharp
(**
---
title: fsi.AddPrinter test
---
*)

type Foo = { X: int }

fsi.AddPrinter(fun (f: Foo) -> sprintf "FOO(%d)" f.X)

let v = { X = 42 }
v
(*** include-it ***)
```

Run:

```bash
dotnet fsdocs build --noapidocs --output output --eval
```

**Observed**: error message `The value, namespace, type or module 'fsi' is not defined` is printed.

**Expected**: no error (or at most a suppressed warning), since `fsi` is a well-known identifier in FSI scripts and `fsi.AddPrinter` is a [documented feature](https://fsprojects.github.io/FSharp.Formatting/evaluation.html#Using-AddPrinter-and-AddHtmlPrinter) of fsdocs literate evaluation.

## Impact

The error is cosmetically confusing but more importantly, in some scenarios (particularly when referencing external DLLs via `#r` and using interface types in the printer), the compile-time error appears to prevent subsequent code from evaluating, causing `include-it` to produce "No value returned by any evaluator" for all following snippets.

## Environment

- fsdocs-tool 22.0.0-alpha.2
- .NET SDK 10.0.104
- F# Interactive 14.0.104.0 (F# 10.0)
- Linux (Ubuntu)

dsyme:
## Description

Fixes #1139

`fsdocs build --eval` reports a spurious error when a literate script uses `fsi.AddPrinter`, `fsi.AddPrintTransformer`, or `fsi.AddHtmlPrinter`:

```
docs/test.fsx: error(8,0)-(8,3) The value, namespace, type or module 'fsi' is not defined.
```

The error comes from the compile-time type-checking pass (not evaluation). The `fsi.AddPrinter` call executes correctly at runtime.

## Root Cause

`FSharpChecker.GetProjectOptionsFromScript` never includes `FSharp.Compiler.Interactive.Settings.dll` in the returned references — regardless of `--noframework`, `useFsiAuxLib`, or `useSdkRefs` flags. Without this reference, `ParseAndCheckFileInProject` cannot resolve the `fsi` identifier.

This was confirmed by a standalone debug program that tested every combination of flags: only an explicit `-r:` reference to the settings DLL resolves the `fsi` identifier at type-check time.

## Fix

- **`YaafFSharpScripting.fs`**: Added `fsiSettingsDll` lazy helper to `FSharpAssemblyHelper` that locates `FSharp.Compiler.Interactive.Settings.dll` by searching `AppContext.BaseDirectory` first, then the SDK's `FSharp/` directory (derived from the shared-framework runtime path).
- **`CodeFormatAgent.fs`**: When building project options for type-checking, adds the settings DLL as an explicit `-r:` reference if found.
- **`EvalTests.fs`**: Added regression test that asserts zero diagnostics and correct printer output when using `fsi.AddPrinter` in a literate script.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [22.0.0-alpha.3] - 2026-04-02
+
 ### Fixed
 * Fix spurious `'fsi' is not defined` error during literate script type-checking when scripts use `fsi.AddPrinter` or related APIs. The `FSharp.Compiler.Interactive.Settings.dll` reference is now explicitly added to the type-checker options. [#1139](https://github.com/fsprojects/FSharp.Formatting/issues/1139)
 * Fix literate script comment parser prematurely closing `(**` blocks when the markdown text contained nested `(*** ... ***)` references (e.g. in backtick-quoted command examples), causing subsequent content to be silently dropped from HTML output.


dsyme:
@nojaf I'll swear there has been an FSharp.Compiler.Service regression here with regard to this:

> FSharpChecker.GetProjectOptionsFromScript never includes FSharp.Compiler.Interactive.Settings.dll in the returned references — regardless of --noframework, useFsiAuxLib, or useSdkRefs flags. Without this reference, ParseAndCheckFileInProject cannot resolve the fsi identifier.

`useFsiAuxLib=true` is always meant to add this reference. 

I notice Ionide is showing `fsi` as unresolved in scripts and this may be related

We shouldn't have to do this kind of resolution ourselves.

dsyme:
@nojaf merging as using the latest version in Deedle and trying to iron out problems

PR #1288, opened 2026-09-07
Rewrite fsdocs watch as a lazy in-memory dev server

nojaf:
## Why this is a draft

This branch bundles a full rewrite of `fsdocs watch` plus a number of things that fell out of it along the way. That is far too much to review or merge as one unit, so it is opened as a **draft** to start the conversation and get early feedback on the direction.

The plan is to land the pieces below **incrementally**, as separate, independently reviewable PRs. Think of this PR as the map, not the merge candidate.

## The core design: one incremental dependency graph

Everything else here follows from this, so it is worth reading first.

The dev server models the whole site as an **incremental dependency graph** built with [FSharp.Data.Adaptive](https://github.com/fsprojects/FSharp.Data.Adaptive). Every URL (a content page, an API page, the search index, `llms.txt`) is a node that is computed the first time somebody asks for it and then cached. The inputs of that node are also nodes: the file itself, the files it `#load`s, the templates that were tried, the navigation, the substitutions, the cracked projects, the API model. So the graph knows exactly which pages a given file can influence, and nothing else gets touched.

Concretely:

* Watched files live in an adaptive map of *stamps* (length, last write time, and a content hash for the files whose content actually feeds the graph). The stamp, not the file system event, is what invalidates.
* There is a single entry point, `Refresh path`, and it is the only way the graph ever learns that something changed. Both the `FileSystemWatcher` and a periodic reconciler (which walks the watched roots every two seconds as a guard against missed OS events) go through it. The request path never stats a file to decide whether a page is fresh.
* Because the hash is part of the stamp, a **byte-identical rewrite invalidates nothing**. Save without changing anything, or let a formatter rewrite a file to the same bytes, and the browser does not reload and no page is rebuilt.
* Recomputation stops as early as it can. `Adaptive.mapCached` at the expensive nodes compares the recomputed value structurally and does not propagate when it is unchanged. Editing a heading in `a.md` rebuilds `a.html` and refreshes the navigation of the other pages, without re-rendering them.

The practical effect: after the first request, edits cost roughly what the edit actually implies and nothing more.

## What is in here

* **Lazy, folder-less dev server.** `fsdocs watch` no longer copies the site into an output folder. Nothing is written to disk: a page is built the first time the browser asks for it, static files are served straight from their source location. This is the big speed-up. Startup no longer means "render every page in the repo before you can look at one of them", and a docs set with hundreds of pages costs the same to start as a docs set with three. We only build what is actually requested, and after that only what actually changed. A page that fails to build returns a `500` with the error while the rest of the site keeps working and the process stays up, even with `--strict`, so a typo in one script does not take the session down. `--output`, `--clean` and `--saveimages` are ignored by `watch`, and `--noserver` is removed, since watching without serving no longer does anything.
* **Quiet by default logging.** The reason for this one is the experience, not the plumbing: the tool used to shout a wall of text at you, one line per file, and the things you needed to see drowned in it. Logging now goes through `Microsoft.Extensions.Logging`. The libraries are silent unless a host opts in (`Logging.UseConsole level`, or your own `LoggerFactory`), and the tool has `-v` / `--verbosity quiet|minimal|normal|detailed|diagnostic` on `build`, `watch`, `convert` and `init`. `normal` prints one line per phase with counts and durations instead of one line per file; warnings and errors go to standard error with a `warning:` / `error:` prefix so they stand out and can be filtered.
* **Diagnostic pages.** `/.fsdocs/doctor` (HTML) and `/.fsdocs/doctor.json` show the cracked projects, the substitutions and where each one came from, the templates that were tried and chosen, the navigation with the source of every title, the routes, the cache state of every URL, the recent file events, the collected errors and the recent log lines. "Why is my title wrong / my logo missing / this page stale" stops being guesswork. The JSON form exists so you can hand the whole state to an LLM.
* **Relative links.** `{{root}}` is now the path from the page to the root of the site (`./`, `../`, ...) instead of the absolute site URL, so a generated site works from any host name, from a sub path, straight off the file system, and the dev server can be reached through any address it is bound to (a reverse proxy, a Codespaces forward). The absolute URL moves to a new `{{fsdocs-site-root}}` substitution, used by the things that genuinely need it (`og:url`, `twitter:site`, `llms.txt`). `--root` on `watch` becomes `--site-root` and only sets that URL. The dev server serves a folder URL from its index page and redirects `/folder` to `/folder/`, as static web servers do. Breaking for custom templates that use `{{root}}` where an absolute URL is required.

  *Does a sub path still work?* Yes, and without any configuration. This is the opposite of Vite's [public base path](https://vite.dev/guide/build#public-base-path): Vite needs `base` because it emits absolute paths, relative links need nothing. Deployed at `https://fsprojects.github.io/fantomas/`, `reference/index.html` links to `../content/fsdocs-default.css`, which the browser resolves against `.../fantomas/reference/`, so the `/fantomas/` part never appears in the HTML and cannot be wrong. The same output works at a domain root, under a sub path, behind a reverse proxy, and from `file://`. It was the old behaviour that needed the Vite style knob, since the deploy URL had to be baked in at build time with `--parameters root`.

  Two things to keep in mind. Relative links depend on folder URLs keeping their trailing slash (`/reference/`, not `/reference`); static hosts redirect for you and the dev server now does the same. And the absolute URL is still needed where it genuinely must be absolute (`og:url`, `twitter:site`, `llms.txt`), which is what `{{fsdocs-site-root}}` is for; it defaults to `<PackageProjectUrl>`, and getting it wrong now costs you a social card rather than every link on the site.
* **Project cracking by evaluation.** MSBuild properties are read with `dotnet msbuild --getProperty`, in parallel, with no design-time build, and test projects and executables are dropped before any expensive work happens. The compiler references come from a design-time build that runs only for the documented projects, only when the API docs are actually generated, and is cached in `.fsdocs/references`. Ionide.ProjInfo is no longer used for the design-time build, only to read solution files. The `.fsdocs/cache` key now covers the discovered project files, `Directory.Build.props`, `Directory.Build.targets`, `Directory.Packages.props` and `global.json`, so editing those no longer serves stale project settings.
* **Design-time refinement of substitutions.** Properties set by MSBuild targets (a `Version` computed from a changelog by `Ionide.KeepAChangelog.Tasks`, for example) are invisible to a plain evaluation. So `build` uses the refined values from the design-time build, and `watch` serves the evaluated values immediately, then rebuilds the affected pages once the background design-time build fills in the real ones. You get a page now and the correct value shortly after, rather than waiting for MSBuild before the first paint. The doctor page shows per project whether the design-time build ran, which substitutions it changed, and has a button to run it again.
* **Background API docs and project watching.** The API reference is built in the background at startup and again after a project DLL or project file changes, so no page request ever waits for it. Project files and the solution-wide MSBuild files are watched too: a change re-cracks the projects, so things like `<FsDocsLogoSource>` and the compiler references are picked up without a restart.
* **`--host`.** `fsdocs watch --host 0.0.0.0` binds the server to all interfaces so the site can be browsed from another machine on the network. This is only reasonable now that links are relative.
* **Lazily fetched search index.** `index.json` is the one URL that needs every page to exist, so it is now fetched when the search dialog is first opened instead of on page load. Without this, the lazy server would eagerly build the whole site on the first request anyway.
* **The page no longer jumps to the top on reload.** Editing a paragraph halfway down a long page used to send you back to the top on every save, which is the kind of small friction that makes you stop using watch mode. The scroll position is now saved before the live reload and restored on the new page. Getting this to actually feel right took some care: the default template scrolls `<main>` rather than the window, so the saved position belongs to that container; the restore happens on `pagereveal` (before the first paint) so a cross-document view transition animates in at the right position instead of jumping afterwards, with `load` as the fallback for browsers without it; and content can still grow after `load` (Mermaid, for one, only starts fetching then), so a `ResizeObserver` reapplies the target for as long as the content keeps resizing and stops once it has been quiet for a few seconds, the user scrolls themselves, or a generous cap is hit.
* **Smaller fixes found along the way:** fall back to `port + 1` when the port is taken, fix `fsdocs-next-page-link` resolution in the dev server, normalize URLs for bad input (no `//` mid path), warn at startup when the configured logo or favicon is not served, add signature files to `fsdocs-tool`, and clean up analyzer warnings.

## How I would like to proceed

Split this into focused PRs, each with its own tests, roughly in this order: relative links, project cracking, logging, then the dev server rewrite (adaptive graph, doctor page, background API docs) which is the largest piece and depends on the rest. Happy to reorder based on what you would rather see first, or to drop anything that you would rather not have.

## Known follow-up

Relative links have one gap: a `404.html` is served by the host for an arbitrary path (GitHub Pages does this), so its relative links, computed for its own depth, resolve against the requested URL instead. The tool does not generate a 404 page today, but a hand-written `docs/404.md` would render unstyled under a deep URL. That page wants root-absolute links, and is worth doing as a follow-up rather than holding up this work.

@dsyme does this all sound reasonable to you? Happy to go step by step and send smaller PRs to work on this.


dsyme:
This is fantastic and exactly the design driection that should have been pursued originally

I love that we're adopting FSharp.Data.Adaptive in this tool. It's so powerful

dsyme:
/repo-assist review this pull request and reply with a comment added to this PR conversation

dsyme:
Lots of merge conflicts after #1289 , sorry, you will need to re-format with Fantomas 8 then the merge conflicts should mostly disappear.

nojaf:
@dsyme well, we could also live a little and merge these commits in as is.
I rewrote the history for each new concept is nicely in one commit.
Added an 23 alpha entry in the release notes. Best way to test this will be to use it.
Had a great experience with it here locally.

dsyme (approved):

PR #1315, opened 2026-09-10
Skip union case tester properties in API docs

### LINKED ISSUE #1313: Missing-docs warning FD0001 reported for compiler-generated union case IsXxx properties
## Summary

With `FsDocsWarnOnMissingDocs` enabled, fsdocs (23.0.0-alpha.1) reports `FD0001: no documentation` for the compiler-generated `IsXxx` union case tester properties. These properties are synthesized by the F# compiler for every union case and cannot carry XML documentation in source, so the warnings are noise that no source change can fix.

## Repro

Building the API docs for `FSharp.Compiler.Service` from https://github.com/dotnet/fsharp with `<FsDocsWarnOnMissingDocs>true</FsDocsWarnOnMissingDocs>`:

```
src/Compiler/AbstractIL/il.fsi(158,6): warning FD0001: no documentation for 'P:FSharp.Compiler.AbstractIL.IL.ILArgConvention.IsDefault'
src/Compiler/AbstractIL/il.fsi(160,6): warning FD0001: no documentation for 'P:FSharp.Compiler.AbstractIL.IL.ILArgConvention.IsStdCall'
src/Compiler/AbstractIL/il.fsi(161,6): warning FD0001: no documentation for 'P:FSharp.Compiler.AbstractIL.IL.ILArgConvention.IsThisCall'
```

For that one assembly, 1059 of the 2073 `FD0001` warnings are for `.Is<CaseName>` properties on union types, so about half of the missing-docs output is unfixable.

A minimal repro is any documented union:

```fsharp
/// A convention
type Convention =
    /// The default
    | Default
    /// Standard call
    | StdCall
```

fsdocs then warns about `P:...Convention.IsDefault` and `P:...Convention.IsStdCall`.

## Suggestion

Skip the missing-docs warning (and arguably the members altogether, or document them as "true if this is the `X` case") for union case tester properties. They are identifiable as properties named `Is<CaseName>` on a union entity where `<CaseName>` matches one of `entity.UnionCases`, or via `IsUnionCaseTester` on `FSharpMemberOrFunctionOrValue` if FCS exposes it in the version you build against.

Environment: fsdocs-tool 23.0.0-alpha.1, Linux, .NET SDK 11.0.100-rc.1.

nojaf:
The compiler synthesizes an IsCase property for every union case. They cannot carry XML documentation, so they were listed as undocumented members on union pages and FsDocsWarnOnMissingDocs reported FD0001 for each one, about half of all warnings for a union-heavy assembly.

Treat IsUnionCaseTester like IsCompilerGenerated when reading members, so the testers are neither documented nor reported as missing.

@dsyme would like a review from you on this. I think this is fine, but a another set of eyes might be worth it.

Fixes #1313

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 * Fix tooltip not being interactive: moving the mouse from a code token into its tooltip now keeps the tooltip open, allowing users to select and copy the tooltip text. [#949](https://github.com/fsprojects/FSharp.Formatting/issues/949)
+* The compiler-generated `IsCase` union case tester properties are no longer listed as members of a union type, and `FsDocsWarnOnMissingDocs` no longer reports `FD0001` for them: they cannot carry XML documentation. [#1313](https://github.com/fsprojects/FSharp.Formatting/issues/1313)
 * `fsdocs watch` rebuilds a page when a file its script depends on changes, following `#load` transitively and `#r` to local files, wherever those files are (a dot folder, outside the input folder). The directives are read from the syntax tree, so a `#load` in a comment does not count. [#1309](https://github.com/fsprojects/FSharp.Formatting/issues/1309)
 
 ## [23.0.0-alpha.1] - 2026-09-08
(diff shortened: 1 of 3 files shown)


dsyme (approved):

dsyme:
Looks good

